PR #9, opened 2020-05-14
Alum Directory Component

awilli70:
Page 1:
Screen Shot 2020-05-13 at 10 14 27 PM
Page 2:
Screen Shot 2020-05-13 at 10 14 38 PM

### DIFF
--- a/client/src/components/AlumniDirectory.js
+++ b/client/src/components/AlumniDirectory.js
@@ -1,0 +1,107 @@
+import React, { Component } from 'react';
+import { Card, Image, Search, Pagination, Grid, Segment } from 'semantic-ui-react'
+
+/*
+props:
+- entries: dictionary
+- query: string
+*/
+export default class AlumniDirectory extends Component {
+    state = {
+        activePage: 1,
+        numEntries: Object.keys(this.props.entries).length,
+        isLoading: false,
+        value: '',
+        totalPages: Math.ceil(Object.keys(this.props.entries).length/3)
+    }
+
+    handlePaginationChange = (e, { activePage }) => {
+        this.setState({ activePage })
+    }
+    handleSearchChange = (e, { value }) => {
+        this.setState({ value })
+    }
+
+    render(){
+        const {
+            totalPages,
+            activePage
+        } = this.state
+
+        return (
+            <Grid columns={1} divided>
+                <Grid.Row stretched>
+                    <Grid.Column>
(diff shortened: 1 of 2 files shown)


client/src/components/AlumniDirectory.js
```diff
@@ -0,0 +1,93 @@
...
+import { Card, Image, Search, Pagination, Grid, Segment } from 'semantic-ui-react'
+
+/*
+props:
+- entries: list of alumni profiles
+*/
+export default class AlumniDirectory extends Component {
+    state = {
+        activePage: 1,
+        numEntries: Object.keys(this.props.entries).length,
```
reshadbinharun, line 11: Any reason we're calling length on keys instead of the entries array itself?

awilli70: Not anymore - it used to be necessary because entries was a js object, but I switched over to a list.
I'll change it now

client/src/components/AlumniDirectory.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    render(){
+        const {
+            totalPages,
+            activePage
+        } = this.state
+
+        let profiles=[]
+        for (let post in this.props.entries) {
```
reshadbinharun, line 31: Might be cleaner to iterate over items using `let post of this.props.entries` and then using each post entry inside iteration like `post.imageUrl`

client/src/components/AlumniDirectory.js
```diff
@@ -0,0 +1,93 @@
...
+                                open={false}
+                                showNoResults={false}
+                                onSearchChange={this.handleSearchChange}
+                                input={{fluid: true}}
+                                placeholder={"Search"}
+                            />
+                    </Grid.Column>
+                </Grid.Row>
+                
+                {profiles[(activePage - 1) * 3]}
```
reshadbinharun, line 75: I like the idea of setting the page size to 3! can we declare this on the file somewhere? PAGE_SIZE = 3, so we can easily change this if need be!

client/src/components/AlumniDirectory.js
awilli70: fixed!

client/src/components/AlumniDirectory.js
awilli70: pageSize is now a state, used in pageGenerator function

PR #10, opened 2020-05-16
React Redux Store setup

reshadbinharun:
image

* For components that need access to `store` state and/or write to `store`, we will need to add
- mapStateToProps
- mapDispatchToProps
... and wrap in connect(mapStateToProps, mapDispatchToProps)(ComponentClass) to allow it read/write access to store!

We do not necessarily need to use the store for all components.


client/src/App.js
```diff
@@ -20,6 +24,26 @@ export const PATHS = {
   register: "/register",
 }
 
+/*
+  STORE SETUP
+*/
+
+const mapStateToProps = state => {
```
reshadbinharun, line 31: The state should correspond to how `combineReducers` combines reducers, e.g. `countState` is the key that stores the store state change brought about by the `incrementReducer`

client/src/App.js
```diff
@@ -20,6 +24,26 @@ export const PATHS = {
...
+  return ({
+    schoolName: state.schoolName,
+    count: state.countState.count
+  })
+}
+
+const mapDispatchToProps = dispatch => ({
```
reshadbinharun, line 38: A `dispatch` is used to trigger an action. The payload passed in is used to make changes to the store state in a reducer.

client/src/App.js
```diff
@@ -20,6 +24,26 @@ export const PATHS = {
   register: "/register",
 }
 
+/*
```
reshadbinharun, line 27: Each component that needs to access/write to store, needs these mapState methods defined. These methods are passed in as arguments to `connect`

client/src/App.js
```diff
@@ -190,14 +224,32 @@ export default class App extends Component {
...
           />
+          <Route exact path={`/playground`} render={(props) => 
+              <>
+                <Navbar
+                    navItems={alumniNavBarItems()}
+                    activeItem={'playground'}
+                />
+                <Button
+                  primary
+                  onClick={(e) => this.props.increment(1)}
```
reshadbinharun, line 243: dispatch methods defined in `mapDispatchToProps` are available as props.

client/src/App.js
```diff
@@ -190,14 +224,32 @@ export default class App extends Component {
...
+                >
+                  Click to increase counter
+                </Button>
+                <Segment>
+                  Count in store is {this.props.count ? this.props.count : 0}
```
reshadbinharun, line 248: data defined in `mapStateToProps` is available as props.

client/src/App.js
```diff
@@ -267,3 +319,5 @@ export default class App extends Component {
     )
   }
 }
+
+export default connect(mapStateToProps, mapDispatchToProps)(App);
```
reshadbinharun, line 323: The component must be connected in this manner, to allow access to state.

client/src/index.js
```diff
@@ -1,8 +1,16 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import { Provider } from 'react-redux';
+import configureStore from "./redux/store"
 
 ReactDOM.render(
-        <App/>
+        <Provider
```
reshadbinharun, line 8: The Provider creates a store, and subscribes all nested children to the store.

client/src/redux/actions.js
```diff
@@ -0,0 +1,13 @@
+export const testAction = () => dispatch => {
```
reshadbinharun, line 1: Actions are plain javascript objects. These methods are `action creators` in that they return actions. Actions are intended to represent a change that should happen to the store. They don't actually mutate the store in any way!

client/src/redux/actions.js
```diff
@@ -0,0 +1,13 @@
+export const testAction = () => {
```
reshadbinharun, line 1: Actions are plain javascript objects. These methods are action creators in that they return actions. Actions are intended to represent a change that should happen to the store. They don't actually mutate the store in any way!

client/src/redux/reducers.js
```diff
@@ -0,0 +1,27 @@
+import { combineReducers } from 'redux';
```
reshadbinharun, line 1: Reducers cause the actual mutations to the store. All reducers will `listen` for all actions, but only mutate for the action it is concerned with.

client/src/redux/reducers.js
```diff
@@ -0,0 +1,27 @@
...
+        case 'INCREMENT_ACTION':
+            return {
+                count: (state.count ? state.count : 0) + action.payload
+            }
+            default:
+                return state
+        }
+}
+
+export default combineReducers({
```
reshadbinharun, line 24: The way combineReducers builds this object is import for `mapStateToProps`

awilli70:
Looks awesome! Thanks for the documentation

PR #14, opened 2020-05-19
created request model, endpoint, utils

awilli70:
Models
requestSchema:
student - studentSchema._id
alumni - alumniSchema._id
time - string, required
zoomLink - string, required (Note: this means an alumni must have a zoomLink before they can be requested) - maybe have a form that pops up upon first login?
topic - string, required
status - string enum ['Awaiting Confirmation',
                                    'Completed',
                                    'Rejected',
                                    'Feedback Provided']
(The string must match one of these categories or it will fail validation)
intro - string
notes - string

Endpoints
POST /request/addRequest - adds a request to the Requests collection
body: student_email, alumni_email, time, topic, status, intro, note
(emails will be the easiest way to identify users - because the alumni is an object on the frontend, we will have access to its email even if the user won't)

GET /mongoose-util/allRequests - gets all requests from Requests collection
GET /mongoose-util/populateAllRequests - gets all requests with populated student and alumni (instead of just their ._id)
GET /mongoose-util/data/clear/requests - deletes all requests from Requests

modified:
GET /mongoose-util/data/clear/all - now includes requests (previously just alumni, student, user)

sidenote: currently, studentsRouter does not exist - this means all code in app.js does not work after app.use(/students/). options are removing its app.use, or including studentsRouter

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -109,9 +109,7 @@ async function main() {
       next();
     }, mongooseUtilRouter);
 
-    app.use('/request/', (req, res, next) => {
-      next();
-    }, requestRouter);
+    app.use('/request/', requestRouter);
 
     app.use('/students/', (req, res, next) => {
       next();
--- a/server/app.js
+++ b/server/app.js
@@ -14,6 +14,8 @@ var JWTStrategy = require("passport-jwt").Strategy;
 var indexRouter = require('./routes/index');
 var utilRouter = require('./routes/util');
 var mongooseUtilRouter = require('./routes/utilMongoose');
+var requestRouter = require('./routes/requests');
+
 require('dotenv').config();
 
 var app = express();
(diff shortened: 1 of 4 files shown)


server/app.js
```diff
@@ -107,6 +109,10 @@ async function main() {
       next();
     }, mongooseUtilRouter);
 
+    app.use('/request/', (req, res, next) => {
```
reshadbinharun, line 112: Since we are not adding middleware here, this can just be `app.use('/request/'. requestRouter)`

server/models/requestSchema.js
```diff
@@ -0,0 +1,24 @@
...
+
+const requestSchema = new Schema(
+  {
+    student: {type: Schema.Types.ObjectId, ref: 'Student'},
+    alumni: {type: Schema.Types.ObjectId, ref: 'Alumni'},
+    time: {type: String, required: true},
+    zoomLink: {type: String},
+    topic: {type: String, required: true},
+    status: {type: String, 
+                enum: [ 'Awaiting Confirmation',
```
reshadbinharun, line 13: Forgot to add this in DataModel: `Confirmed`

reshadbinharun: so 5 possible states

PR #17, opened 2020-05-22
Use LinkedIn API to get images for users

reshadbinharun:
* LinkedIn API (r_liteprofile permission) allows getting firstName, lastName, and profile picture

This PR sets up to get public link for LinkedIn profile photo URL and store it on the alumni/student record


client/src/components/Profile.js
```diff
@@ -22,9 +23,7 @@ export default class Profile extends Component {
         const isViewOnly = this.props.isViewOnly;
 
         const linkedInUpdate = (
-            <Button primary>
-                Update info with LinkedIn
-            </Button>
+            <LinkedInUpdate/>
```
reshadbinharun, line 26: TODO: email should be passed in as prop here if we do in fact plan to use this method of fetching profile pics

client/src/components/Signup.js
```diff
@@ -184,13 +184,17 @@ export default class Signup extends React.Component {
             e.preventDefault();
             const endPoint = this.props.isAlumni ? '/alumni' : '/student';
             try {
-                const result = makeCall(payload, endPoint, 'post')
+                const result = await makeCall(payload, endPoint, 'post')
```
reshadbinharun, line 187: Bonus fixes!

reshadbinharun:
We should discuss if it's worth using the LinkedIn API at all!
Also (https://thenextweb.com/dd/2015/02/12/linkedin-takes-aim-developers-plans-lock-apis/)[Boo!]

client/src/components/Profile.js
awilli70: Another option would be to just pass in the profile data to this component and store it as a state - we would still need to have isViewOnly

awilli70:
LinkedIn at this point seems to have very limited utility for the project as a whole - if its sole purpose in the application is just to grab a profile picture url, it offers very little convenience for the increase in code and means we rely on an API that seems pretty likely to change given the amount of features they have already closed off

reshadbinharun:
> LinkedIn at this point seems to have very limited utility for the project as a whole - if its sole purpose in the application is just to grab a profile picture url, it offers very little convenience for the increase in code and means we rely on an API that seems pretty likely to change given the amount of features they have already closed off
> 
> We can talk more about this on Saturday!

Agreed! We can limit our potential use to only grabbing the profile photo, and have this end-point available should we choose to use it!

PR #21, opened 2020-05-23
Cookies! and StudentRoutes

awilli70:
#### Backend:
new routes for users: user.js
added userRoutes to app.js

Modified: /login - the cookie is now created once the profile details have been gathered - this has changed nothing behavior wise but allows the cookie to contain more useful info

### Get /alumni,students,user/one/
body: 
- email: string
returns matching object

#### Frontend:
LoginForm is unchanged from its original form
App.js now has student routes
//TODO: Fix profile routes - currently broken because profile and role aren't props
// This is ok for now because they will be overhauled soon
Default route switched from null to Home - this makes the transition from login to home less jarring
App.js fetches role and profile from the cookie! - if user is logged with valid token, the jwt will be parsed and used to set states
logout now clears local storage - this will be useful if we ever put anything in local storage in the future! 

ComponentWillMount now asynchronous to eliminate occasional issues with makeCall
PersistedState was not doing anything - removed

liftPayload now automatically refreshes page
ezgif-3-6944129ff338

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -151,7 +151,7 @@ class App extends Component {
   async componentWillMount() {
     var role;
     var profile;
-    var email;
+    var id;
     this.setState({
       fetchingAuth: true
     }) 
@@ -164,8 +164,8 @@ class App extends Component {
         var jwtVal = document.cookie.replace(/(?:(?:^|.*;\s*)jwt\s*\=\s*([^;]*).*$)|^.*$/, "$1");
         const parsedJWT = JSON.parse(atob(jwtVal.split('.')[1]));
         role = parsedJWT.role;
-        email = parsedJWT.email;
-        profile = await this.fetchProfile(role, email);
+        id = parsedJWT.id;
+        profile = await this.fetchProfile(role, id);
         this.setState({
           role: role,
           userDetails: profile,
@@ -186,12 +186,12 @@ class App extends Component {
     }
   }
 
-  async fetchProfile(role, email) {
+  async fetchProfile(role, id) {
     let result;
     if (role === 'STUDENT') {
-      result = await makeCall({email: email}, '/student/one', 'post')
+      result = await makeCall({}, ('/student/'+id), 'get')
     } else {
-      result = await makeCall({email: email}, '/alumni/one', 'post')
+      result = await makeCall({}, ('/alumni/'+id), 'get')
     }
     return result.result
   }
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -49,29 +49,29 @@ const mapDispatchToProps = dispatch => ({
 */
 
 // TODO extract Routes into this higher-order component
-function withLoginCheck(RouterComponent, isLoggedIn, navItems, routePath, activeItem) {
-  return class extends React.Component {
-    constructor(props) {
-      super(props)
-    }
-    render() {
-      return (
-        <Route exact path={routePath} render={(props) => 
-            isLoggedIn ? 
-            <>
-                <Navbar
-                    navItems={navItems}
-                    activeItem={activeItem}
-                />
-                <RouterComponent {...this.props} />
-            </> :
-            <Redirect to={'/login'} />
-          }
-        />
-      )
-    }
-  }
-}
+// function withLoginCheck(RouterComponent, isLoggedIn, navItems, routePath, activeItem) {
+//   return class extends React.Component {
+//     constructor(props) {
+//       super(props)
+//     }
+//     render() {
+//       return (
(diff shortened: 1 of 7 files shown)


awilli70:
I've been looking into how to read a cookie's value on the frontend - I'm going to go back and do a second pass later because currently this is VERY insecure

It will involve a little more backend work re: cookie creation, but should result in a much safer way to persist states without using localstorage

awilli70:
Now uses the cookie's jwt to persist the value of role and details!
This is awesome, because:
- login and persisting profile info is all in one place (the cookie) so it will always be available at the same time
- It's WAY more secure. Users cannot spoof one another or edit their permissions, which they could do with local storage

Behavior is exactly the same

client/src/App.js
```diff
@@ -122,42 +148,39 @@ class App extends Component {
...
-    }, async () => {
-      try {
-        const result = await makeCall({}, '/isLoggedIn', 'get')
+    }) 
+    try {
+      const result = await makeCall({}, '/isLoggedIn', 'get')
+      this.setState({
+        fetchingAuth: false
+      });
+      if (result && !result.error) {
```
awilli70, line 162: This is where the JWT is decoded and used to set the state
Regex lifted from stackoverflow, as is the parsing

client/src/App.js
```diff
@@ -122,42 +148,39 @@ class App extends Component {
...
-        }
-      } catch (e) {
-          this.setState({
-            loggedIn: false,
-            fetchingAuth: false,
-          });
-          console.log("Error: App#componentDidMount", e)
-      }
-    })
-    const persistedState = JSON.parse(localStorage.getItem(App_LS));
```
awilli70, line 152: Persisted state was not being used due to quirks with setState in liftedPayload, and is now unnecessary

reshadbinharun: Good riddance! Let's try to get rid of using localStorage all together!

client/src/App.js
```diff
@@ -168,6 +191,7 @@ class App extends Component {
   }
 
   async logout() {
+    await localStorage.clear()
```
reshadbinharun, line 194: May be unnecessary now!

server/routes/index.js
```diff
@@ -39,21 +39,23 @@ router.post('/login', (req, res, next) => {
...
           }
-          const cookie = jwt.sign(JSON.stringify(payload), JWT_SECRET);
-          // set jwt-signed cookie on response
-          res.cookie('jwt', cookie);
           try {
             let userRole = user.role && user.role.toUpperCase()
             if (userRole === "ALUMNI") {
               const alumni = await alumniSchema.findOne({email: user.email});
+              payload.details = alumni
+              const cookie = jwt.sign(JSON.stringify(payload), JWT_SECRET);
```
reshadbinharun, line 56: Let's sign the cookie with `email` and `role` payload, and extract that information from the cookie! The components can use that information to fetch other details as needed!

awilli70:
This is what the JWT value looks like now
Profile info is fetched in a new helper function

{
  "email": "email@email.com",
  "role": "ALUMNI",
  "expires": 1590361131043
}

server/routes/students.js
```diff
@@ -55,4 +55,14 @@ router.post('/', async (req, res, next) => {
     }
 });
 
+router.post('/one', async (req, res, next) => {
```
reshadbinharun, line 58: Can we change this method to `GET` and have the path be `/:email` so that we can read `email` from `req.params.email`

server/routes/user.js
```diff
@@ -0,0 +1,16 @@
+var express = require('express');
+var router = express.Router();
+var userSchema = require('../models/userSchema');
+require('mongoose').Promise = global.Promise
+
+router.post('/one', async (req, res, next) => {
```
reshadbinharun, line 6: same here!

server/routes/alumni.js
```diff
@@ -78,4 +78,14 @@ router.get('/all', async (req, res, next) => {
     }
 });
 
+router.post('/one', async (req, res, next) => {
```
reshadbinharun, line 81: and here!

server/routes/students.js
awilli70: I hesitate to do this because it leaves PII (the email) in the logs on the client and server side in plaintext, whereas `POST` does not

This is outlined [here](https://security.stackexchange.com/a/29605), and can be seen in the dev console.

This is with `POST`:
Screen Shot 2020-05-24 at 6 16 21 PM

This is with `GET`:
Screen Shot 2020-05-24 at 6 17 01 PM

PR #23, opened 2020-05-25
Frontend/alum profile

awilli70:
Backend:
/util/seed/ now gives students an imageURL

Frontend:
Profile now renamed AlumniProfile
StudentProfile created - students don't have much information to display, so this looks a little bare:
Screen Shot 2020-05-25 at 1 05 46 PM

App.js now uses the renamed components, and passes in only two props (the profile and isViewOnly)

### DIFF
--- a/client/src/components/AlumniProfile.js
+++ b/client/src/components/AlumniProfile.js
@@ -11,6 +11,7 @@ props:
     - location: string
     - company: string
     - jobTitle: string
+    - email: string
 - isViewOnly: bool
 */
 export default class AlumniProfile extends Component {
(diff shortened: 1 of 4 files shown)


client/src/components/AlumniProfile.js
```diff
@@ -4,22 +4,19 @@ import LinkedInUpdate from "./LinkedInUpdate";
...
-        const imageURL = this.props.imageURL;
-        const name = this.props.name;
-        const college = this.props.college;
-        const location = this.props.location;
-        const company = this.props.company;
-        const jobTitle = this.props.jobTitle;
+        const details = this.props.details;
         const isViewOnly = this.props.isViewOnly;
 
         const linkedInUpdate = (
```
reshadbinharun, line 22: Let's add the `email` as prop to `LinkedInUpdate` here and for the StudentProfile!

PR #24, opened 2020-05-25
Backend Timezone adjustment helper methods and unit tests

reshadbinharun:
image

* Added timezoneHelpers module that takes strips or applies timezone to an array of timeSlots. timeSlots have shape
```
  {
    id: 'Sunday-400',
    day: 'Sunday',
    time: 400
  }
```

### DIFF
--- a/server/models/alumniSchema.js
+++ b/server/models/alumniSchema.js
@@ -2,6 +2,21 @@ var mongoose = require('mongoose')
 
 const Schema = mongoose.Schema;
 
+/*
+  All Time Availabilities are to be stored in GMT+0 timezone
+  Example record
+  {
+    id: 'Sunday-400',
+    day: 'Sunday',
+    time: 400
+  }
+*/
+const timeAvailabilitySchema = new Schema({
+  id: {type: String, required: true},
+  day: {type: String, enum: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'], required: true },
+  time: {type: Number, required: true}
+})
+
 const alumniSchema = new Schema(
   {
     imageURL: {type: String, default:'https://cdn.pixabay.com/photo/2015/10/05/22/37/blank-profile-picture-973460_1280.png'},
@@ -14,7 +29,7 @@ const alumniSchema = new Schema(
     college: {type: String, required: false},
     //requests: [{type: Schema.Types.ObjectId, ref: 'requestSchema'}]
     //posts: [{type: Schema.Types.ObjectId, ref: 'postSchema'}]
-    availabilities: [Date],
+    availabilities: [timeAvailabilitySchema],
     timeZone: {type: Number, required: false}, // value for Date.getTimezoneOffset (Daylight Savings prevents this from being constant)
     zoomLink: {type: String, required: false}
   }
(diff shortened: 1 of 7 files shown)


server/models/alumniSchema.js
```diff
@@ -2,6 +2,21 @@ var mongoose = require('mongoose')
...
+/*
+  All Time Availabilities are to be stored in GMT+0 timezone
+  Example record
+  {
+    id: 'Sunday-400',
+    day: 'Sunday',
+    time: 400
+  }
+*/
+const timeAvailabilitySchema = new Schema({
```
awilli70, line 14: Is there a reason why this needs to be a Schema rather than a js object?
A schema is a mongoose datatype used to store objects in a collection; unlike requests, availabilities only interacts with one model so it doesn't need to be stored separately

awilli70: Best practices for schemas are also to give each one a separate file so that there isn't additional hunting required - this would probably also be a good idea for timeAvailability

server/models/alumniSchema.js
```diff
@@ -14,7 +29,7 @@ const alumniSchema = new Schema(
     college: {type: String, required: false},
     //requests: [{type: Schema.Types.ObjectId, ref: 'requestSchema'}]
     //posts: [{type: Schema.Types.ObjectId, ref: 'postSchema'}]
-    availabilities: [Date],
+    availabilities: [timeAvailabilitySchema],
```
awilli70, line 32: `availabilites: [timeAvailabilitySchema]` probably won't work they way you want it to work.

If you want it to be a schema, it is easiest to store it as `availabilities: [{type: Schema.Types.ObjectId, ref: 'timeAvailabilitySchema'}]` - but I'm still not sure it makes sense to store availabilities elsewhere

As an object, it could just be `availabilities: [{type: Object, required: false}]

server/models/alumniSchema.js
```diff
@@ -14,7 +29,7 @@ const alumniSchema = new Schema(
...
     timeZone: {type: Number, required: false}, // value for Date.getTimezoneOffset (Daylight Savings prevents this from being constant)
     zoomLink: {type: String, required: false}
   }
```
awilli70, line 35: timeAvailability isn't being exported

server/routes/alumni.js
```diff
@@ -88,4 +89,17 @@ router.get('/:id', async (req, res, next) => {
     }
 });
 
+router.patch('/timePreferences/:id', async (req, res, next) => {
+    try {
+        const alumni = await alumniSchema.findOne({_id: req.body.id})
+        const timezoneAgnosticPreferences = timezoneHelpers.stripTimezone(req.body.timePreferences, alumni.timeZone)
+        alumni.timePreferences = timezoneAgnosticPreferences
```
awilli70, line 96: timePreferences should be availabilities

server/models/alumniSchema.js
awilli70: This uses subdocuments, a newer feature not used as frequently in tutorials

awilli70:
Only necessary fix is alumni.timePreferences to alumni.availabilities - otherwise looks good!
I have a couple questions about the way availabilities is being stored, but that's more preference than anything else

PR #26, opened 2020-05-28
Alumni can add topics they want to consult and update their zoom link

reshadbinharun:
image
image

* Added `topics` on alumniSchema that stores a list of topics alumni can choose to consult
* Added a separate modal that allows alumni to pick topics they want to consult
* On successful addition of new topics and new time preferences, refresh the page since a refetch is required to show updated times and topics
* Added button to allow alumni to update ZoomLink (I did not tag on a refresh on updating Zoom, since this it's unlikely that the user won't refresh before they need to update the zoom link again, let me know what you think)

### DIFF
--- a/client/src/components/TimePreferencesModal.js
+++ b/client/src/components/TimePreferencesModal.js
@@ -182,8 +182,7 @@ export default class TimePreferencesModal extends Component {
             submitting: true
         }, async () => {
             let payload = {
-                timePreferences: this.state.selectedTimes,
-                email: this.props.email
+                timePreferences: this.state.selectedTimes
             }
             try {
                 // TODO: Need to add end-point
@@ -208,6 +207,7 @@ export default class TimePreferencesModal extends Component {
                             icon: "success",
                         }).then(() => {
                             this.props.closeModal();
+                            window.location.reload();
                         })
                     })
                     
--- a/client/src/components/TimePreferencesModal.js
+++ b/client/src/components/TimePreferencesModal.js
@@ -185,7 +185,6 @@ export default class TimePreferencesModal extends Component {
                 timePreferences: this.state.selectedTimes
             }
             try {
-                // TODO: Need to add end-point
                 const result = await makeCall(payload, `/alumni/timePreferences/${this.props.id}`, 'patch')
                 if (!result || result.error) {
                     this.setState({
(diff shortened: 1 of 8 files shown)


client/src/components/TimePreferencesModal.js
```diff
@@ -208,6 +207,7 @@ export default class TimePreferencesModal extends Component {
                             icon: "success",
                         }).then(() => {
                             this.props.closeModal();
+                            window.location.reload();
```
reshadbinharun, line 210: Added a window.location.reload whenever new topcis and times are successfully added, to cause a refetch of profile data with updated time/topic preferences. Ideally, we can refactor this to not do a refresh, let me know what you think.

server/routes/alumni.js
```diff
@@ -79,22 +79,27 @@ router.get('/all/:timezone', async (req, res, next) => {
...
         res.json({'alumni' : alumni});
     } catch (e) {
-        console.log("Error: util#allAlumni", e);
+        console.log("Error: alumni#allAlumni", e);
         res.status(500).send({'error' : e});
     }
 });
 
-router.get('/:id', async (req, res, next) => {
+router.get('/one/:id', async (req, res, next) => {
```
reshadbinharun, line 87: Put `/one` back in since this was the first part of any `GET` route was getting parsed as `id`!

client/src/components/TimePreferencesModal.js
awilli70: I'm assuming that this is intended to refresh the parent component (AlumniProfile)'s profile data from the backend, allowing the modal to open with updated props

This gives us a couple of options, because we have to make a call to the backend either way
- We can switch from giving availabilities as a prop to letting the modal get the availabilities from the backend using id passed as a prop. This gives extra independence to the preferences modal but means the profile page won't always reflect the changes. This can be done with a ComponentWillMount because the Profile will unmount the component upon submission
- We can set fetch the profile's data in this.closeTimePreferencesModal in AlumniProfile before setting the state of the modal to closed. This means that it is updated immediately upon  submission, and if we ever decide to display preferences on the profile we could update the page. In terms of work, this is probably a lot cheaper and will be a pattern we can reuse later for updating the profile in general

server/routes/alumni.js
awilli70: 👍

client/src/components/TimePreferencesModal.js
reshadbinharun: I really like the second option! Good call!

PR #27, opened 2020-05-29
Frontend/request modal

awilli70:
This PR is mostly frontend changes, though it contains minor tweaks to how timezones work.
Requests will require a bit of an overhaul on the backend, so this serves to get the ball rolling
Ideally the changes to requests will not take too long, and submission can be implemented.

Question: Should how do we want to handle multiple requests to the same timeslot? Do we want to give requests a 1/2 week expiry, and block others from making requests during that same time?

Modal upon clicking Request:
Screen Shot 2020-06-03 at 10 28 12 PM

Searchable Dropdown for availabilities:
Screen Shot 2020-06-03 at 10 29 25 PM

Regular Dropdown for topics:
Screen Shot 2020-06-03 at 10 29 34 PM

Submit Request is only enabled when availability and topic are selected:
Screen Shot 2020-06-03 at 10 29 43 PM

Text area and corresponding states in action:
Screen Shot 2020-06-03 at 10 30 38 PM

### DIFF
--- a/client/src/components/AlumniDirectory.js
+++ b/client/src/components/AlumniDirectory.js
@@ -178,7 +178,7 @@ export default class AlumniDirectory extends Component {
     }
 
     getEntries() {
-        return makeCall(null, `/alumni/all/${this.props.timezone}`, 'get').catch(e => console.log(e))
+        return makeCall(null, `/alumni/all/`, 'get').catch(e => console.log(e))
     }
 
     search(value) {
--- a/client/src/components/AlumniDirectory.js
+++ b/client/src/components/AlumniDirectory.js
@@ -197,7 +197,7 @@ export default class AlumniDirectory extends Component {
             } else {
                 if (this.state.allText[i].match(searchPattern) !== null) {
                     numResults += 1
-                    display.push(this.constructProfile(this.state.entries[i]), i)
+                    display.push(this.constructProfile(this.state.entries[i], i))
                 }
             }
             i++;
--- a/client/src/components/AlumniDirectory.js
+++ b/client/src/components/AlumniDirectory.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { Card, Image, Search, Pagination, Grid, Segment, Button, Dropdown } from 'semantic-ui-react'
 import { makeCall } from '../apis';
+import RequestModal from './RequestModal'
 
 // Filter dropdown options
 const searchOptions = [
@@ -49,18 +50,36 @@ props:
 - timezone: Number
 */
 export default class AlumniDirectory extends Component {
-    state = {
-        activePage: 1,
-        numEntries: 0,
-        isLoading: false,
-        value: '',
-        totalPages: 0,
-        entries:[],
-        gradYears: [],
-        allText: [],
-        display: [],
-        numResults: 0,
-        filter: 'all'
+    constructor(props) {
+        super(props)
+        this.state = {
+            activePage: 1,
+            numEntries: 0,
+            isLoading: false,
+            value: '',
+            totalPages: 0,
+            entries:[],
+            gradYears: [],
+            allText: [],
+            display: [],
+            numResults: 0,
+            filter: 'all',
+            requestModalOpen: false,
+            alumniDetails: null,
+        }
+        this.openRequestModal = this.openRequestModal.bind(this)
+        this.closeRequestModal = this.closeRequestModal.bind(this)
+    }
+
+    closeRequestModal() {
+        this.setState({
+            requestModalOpen: false
+        })
+    }
+    openRequestModal() {
+        this.setState({
+            requestModalOpen: true
+        })
     }
 
     async componentWillMount() {
@@ -77,6 +96,7 @@ export default class AlumniDirectory extends Component {
         let gradYears = [];
         let allText = [];
         let display = [];
+        let i = 0;
         for (let post of entries) {
             if(!gradYears.find(year => year['value'] === post.gradYear)) {
                 gradYears.push({
@@ -91,7 +111,8 @@ export default class AlumniDirectory extends Component {
                          + post.company + ' '
                          + post.name + ' '
                          + post.gradYear);
-            display.push(this.constructProfile(post));
+            display.push(this.constructProfile(post, i));
+            i++;
         }
         gradYears.sort(function(a,b){return a.value-b.value})
         this.setState({
@@ -101,7 +122,7 @@ export default class AlumniDirectory extends Component {
                      })
     }
 
-    constructProfile(post) {
+    constructProfile(post, i) {
         return (
             <Grid.Row columns={2}>
                 <Grid.Column width={4}>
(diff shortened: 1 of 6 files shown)


client/src/components/AlumniDirectory.js
```diff
@@ -197,7 +197,7 @@ export default class AlumniDirectory extends Component {
             } else {
                 if (this.state.allText[i].match(searchPattern) !== null) {
                     numResults += 1
-                    display.push(this.constructProfile(this.state.entries[i]), i)
+                    display.push(this.constructProfile(this.state.entries[i], i))
```
awilli70, line 200: visual bug now fixed

client/src/components/RequestModal.js
```diff
@@ -147,42 +147,39 @@ export default class RequestModal extends Component {
...
-                    days.push({
-                        key: day.day.toLowerCase(),
-                        text: day.day,
-                        value: day.day
-                    })
-                }
-            }
-        }
-        this.setState({dayOptions: days})
+    async createOptions(availabilities) {
```
awilli70, line 165: Here is the bulk of the playing nice explained - the offset returned by JS getTimeOffset makes no sense whatsoever with our date model, and results in weird behavior - basically, I needed to make sure that 

A) Timezones were being stored relative to UTC (they weren't, but only because of offset weirdness) - FIXED
B) Because user timezones were being stored improperly, this caused a lot of otherwise functional code to break - this resulted in weirdness where after saving a preference and refreshing, the time shown would be completely different for alumni, and availabilities would be bizarre in the request modal
C) This meant weirdness when passing student timezone as a prop, which would basically double the offset in the wrong direction.

In short - it now works
I do have a question, which I will bring up during standup -
The applyTimezone in all/:timezone doesn't seem to be doing anything - this is fine, but performance wise it doesn't make sense to fix the timezone for all alumni when only a few will be requested in one session

The payload can be changed to remove the math, but to do so we will have to pass the student/alumni as a prop from App.js - that's fine, but it would also be a good first opportunity to use the store

server/routes/alumni.js
```diff
@@ -41,7 +41,7 @@ router.post('/', async (req, res, next) => {
                 //requests: [{type: Schema.Types.ObjectId, ref: 'requestSchema'}]
                 //posts: [{type: Schema.Types.ObjectId, ref: 'postSchema'}]
                 availabilities: availabilities,
-                timeZone: timeZone,
+                timeZone: -timeZone,
```
awilli70, line 44: Ironically this was all that was needed to get most things to play nice

server/routes/students.js
```diff
@@ -28,7 +28,7 @@ router.post('/', async (req, res, next) => {
                 grade: grade,
                 // requests: [{type: Schema.Types.ObjectId, ref: 'requestSchema'}]
                 // issuesLiked: [{type: Schema.Types.ObjectId, ref: 'issueSchema'}]
-                timeZone: timeZone,
+                timeZone: -timeZone,
```
awilli70, line 31: same here

server/routes/requests.js
```diff
@@ -11,6 +12,22 @@ router.get('/', async (req, res, next) => {
     })
 })
 
+router.patch('/applyRequesterTimezone', async (req, res, next) => {
```
awilli70, line 15: This isn't necessary if we change how /alumni/all/:timezone works BUT it probably makes more sense to do it this way

client/src/components/AlumniDirectory.js
```diff
@@ -49,18 +50,36 @@ props:
...
+            numResults: 0,
+            filter: 'all',
+            requestModalOpen: false,
+            alumniDetails: null,
+        }
+        this.openRequestModal = this.openRequestModal.bind(this)
+        this.closeRequestModal = this.closeRequestModal.bind(this)
+    }
+
+    closeRequestModal() {
```
reshadbinharun, line 74: nit: can be changed to use one handler `toggleRequestModalOpen` that uses `setState({requestModalOpen: !this.state.requestModalOpen})`

client/src/components/AlumniDirectory.js
```diff
@@ -183,6 +221,11 @@ export default class AlumniDirectory extends Component {
         }
         this.setState({ [name]: value })
     }
+    handleRequestButton(e) {
+        this.setState({alumniDetails: this.state.entries[e.currentTarget.dataset.id]})
+        console.log(this.state.alumniDetails)
```
reshadbinharun, line 226: Remove `console.log`

client/src/components/AlumniDirectory.js
```diff
@@ -267,6 +310,14 @@ export default class AlumniDirectory extends Component {
                 
                 {searchRow}
                 {resultsRow}
+                {this.state.requestModalOpen &&
```
reshadbinharun, line 313: Do we need the `this.state.requestModalOpen && `?

client/src/components/RequestModal.js
```diff
@@ -0,0 +1,226 @@
...
+    async createAvailabilityOptions(availabilities) {
+        console.log(availabilities)
+        /* 
+         * Offset is in minutes, annoyingly, and in the opposite direction 
+         * that you'd expect. This means that if you are in UTC -4 (EST), 
+         * it returns 240 (not -240 to reflect being 4 hours behind)
+         * That's why there's a tiny bit of math in the payload, to conform
+         * with our date model (which makes more sense)
+         */
+        let timeOffset = await new Date().getTimezoneOffset()
```
reshadbinharun, line 74: Do we need the `await` here?

client/src/components/RequestModal.js
```diff
@@ -0,0 +1,226 @@
...
+
+    async componentWillMount() {
+        await this.setState({alumni: this.props.alumni})
+        this.createAvailabilityOptions(this.state.alumni.availabilities)
+        this.createTopicOptions(this.state.alumni.topics)
+    }
+
+    async createAvailabilityOptions(availabilities) {
+        console.log(availabilities)
+        /*
```
reshadbinharun, line 67: Thank you for documenting :)

client/src/components/RequestModal.js
```diff
@@ -0,0 +1,226 @@
...
+        console.log(availabilities)
+        /* 
+         * Offset is in minutes, annoyingly, and in the opposite direction 
+         * that you'd expect. This means that if you are in UTC -4 (EST), 
+         * it returns 240 (not -240 to reflect being 4 hours behind)
+         * That's why there's a tiny bit of math in the payload, to conform
+         * with our date model (which makes more sense)
+         */
+        let timeOffset = await new Date().getTimezoneOffset()
+        console.log(timeOffset)
```
reshadbinharun, line 75: Remove `console.logs`

server/routes/alumni.js
reshadbinharun, line 44: So does this need to be `timeZone * -1`? (Not sure)

server/routes/requests.js
```diff
@@ -11,6 +12,22 @@ router.get('/', async (req, res, next) => {
...
+    try {
+        const agnosticAvailabilities = req.body.availabilities;
+        const timezone = req.body.offset
+        const availabilities = await timezoneHelpers.applyTimezone(agnosticAvailabilities, timezone)
+        res.status(200).send({
+            availabilities: availabilities
+        })
+    } catch (e) {
+        console.log(e)
```
reshadbinharun, line 24: nit: May be we can include information about which route failed in `console.error`

reshadbinharun:
Looks solid! A few comments.
As for your questions on request setup, I think we should:
a) Remove availability when an alumnus confirms a call for a given time slot. This might mean that when we fetch availabilities for an alumnus on modal open, we filter out a slot that corresponds to a `confirmed request by the same alumnus`. We can push this off to version 2 if this is too much work, let me know your thoughts
b) I haven't played with the idea of expiry on records in mongo but that does seem like a very powerful tool to keep the system clean! I think we out to set an expiry date of a week since our availabilities are weekly times.

client/src/components/AlumniDirectory.js
awilli70: Yes - otherwise it will try to render without an alumni prop

server/routes/alumni.js
awilli70: nope!

PR #29, opened 2020-06-05
fullstack/request-modal

awilli70:
Updated POST /request/addRequest/
Body:
studentId - pulled from cookie
alumniId - pulled from modal's alumni details
zoomLink - pulled from modal's alumni details
timeId - (Day-Hour) - pulled from modal state
topic - pulled from modal state
note - pulled from modal state
timezone - currently calculated, but will be added to modal state with timezone selection update

Implemented submission in RequestModal to create a request in the DB

Updated requestSchema

### DIFF
--- a/client/src/components/RequestModal.js
+++ b/client/src/components/RequestModal.js
@@ -101,11 +101,26 @@ export default class RequestModal extends Component {
 
     submitRequest(e) {
         e.preventDefault()
+        
         this.setState({
             submitting: true
         }, async () => {
             try {
-                const result = true;
+                const jwtVal = document.cookie.replace(/(?:(?:^|.*;\s*)jwt\s*\=\s*([^;]*).*$)|^.*$/, "$1");
+                const parsedJWT = JSON.parse(atob(jwtVal.split('.')[1]));
+                const requesterId = parsedJWT.id
+                let timeOffset = new Date().getTimezoneOffset()
+                timeOffset = -((timeOffset/60)*100)
+                const payload = {
+                    studentId: requesterId,
+                    alumniId: this.state.alumni._id,
+                    zoomLink: this.state.alumni.zoomLink,
+                    timeId: this.state.availabilityValue,
+                    topic: this.state.topicValue,
+                    note: this.state.note,
+                    timezone: timeOffset
+                }
+                const result = await makeCall(payload, `/request/addRequest/`, 'post')
                 if (!result || result.error) {
                     this.setState({
                         submitting: false
--- a/client/src/components/RequestModal.js
+++ b/client/src/components/RequestModal.js
@@ -34,6 +34,8 @@ export const timeSlotOptions = [
 
 /*
 props:
+    - userDetails: logged in user profile
+    - role: user role
     - modalOpen: boolean
     - closeModal: ()
     - alumni (contains all data from alumni schema)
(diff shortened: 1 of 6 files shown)


client/src/components/RequestModal.js
```diff
@@ -101,11 +101,26 @@ export default class RequestModal extends Component {
...
+        
         this.setState({
             submitting: true
         }, async () => {
             try {
-                const result = true;
+                const jwtVal = document.cookie.replace(/(?:(?:^|.*;\s*)jwt\s*\=\s*([^;]*).*$)|^.*$/, "$1");
+                const parsedJWT = JSON.parse(atob(jwtVal.split('.')[1]));
+                const requesterId = parsedJWT.id
+                let timeOffset = new Date().getTimezoneOffset()
```
awilli70, line 112: This should probably be made a state, but it would only really make sense to do this when we have a timezone dropdown

client/src/components/RequestModal.js
```diff
@@ -101,11 +101,26 @@ export default class RequestModal extends Component {
...
         e.preventDefault()
+        
         this.setState({
             submitting: true
         }, async () => {
             try {
-                const result = true;
+                const jwtVal = document.cookie.replace(/(?:(?:^|.*;\s*)jwt\s*\=\s*([^;]*).*$)|^.*$/, "$1");
+                const parsedJWT = JSON.parse(atob(jwtVal.split('.')[1]));
+                const requesterId = parsedJWT.id
```
reshadbinharun, line 111: Would _id not be present in `userDetails` stored at App level?

client/src/components/RequestModal.js
```diff
@@ -101,11 +101,26 @@ export default class RequestModal extends Component {
...
+                let timeOffset = new Date().getTimezoneOffset()
+                timeOffset = -((timeOffset/60)*100)
+                const payload = {
+                    studentId: requesterId,
```
reshadbinharun, line 115: nit: Can we use `requestor`/`requestee` or `mentee`/`mentee` to accommodate the fact that alumni-alumni requests will also be possible?

server/models/requestSchema.js
```diff
@@ -2,11 +2,19 @@ var mongoose = require('mongoose')
...
+  time: {type: Number, required: true}
+})
+
 const requestSchema = new Schema(
   {
-    student: {type: Schema.Types.ObjectId, ref: 'Student'},
-    alumni: {type: Schema.Types.ObjectId, ref: 'Alumni'},
-    time: {type: String, required: true},
+    requester: {type: String, required: true},
+    requesterObj: {type: Object, required: false},
```
awilli70, line 14: This is never stored when a user makes a request, but allows for the full profile to be returned when we need it (see mongooseUtil)

server/routes/utilMongoose.js
```diff
@@ -359,7 +359,14 @@ router.get('/allRequests', async (req, res, next) => {
...
     try {
-        const dbData = await requestSchema.find().populate('alumni').populate('student')
+        const dbData = await requestSchema.find().populate('mentor')
+        for (let request of dbData) {
+            if (request.requesterRole === 'STUDENT') {
+                request.requesterObj = await studentSchema.findOne({_id: request.requester})
+            } else {
+                request.requesterObj = await alumniSchema.findOne({_id: request.requester})
+            }
+        }
```
awilli70, line 369: We can no longer use a cursor to automatically populate requester profiles because we don't know which collection to reference
This is a workaround. We can still populate the mentor, but for each request we need to determine if they are a student or an alumni, find the correct record, and then return the object.
We can't do this in request.requester, because it will be converted to a string. Instead, I've added the requesterObj as a field used to return the student/alumni object since you can't turn requester into an object

PR #32, opened 2020-06-07
fullstack/requests-view

awilli70:
#### requests.js
- new route: GET request/getRequests/:id/:timeOffset
returns all requests for a given alumni
- new route: PATCH request/updateStatus/:id/:timeOffset
Body: requestId, newStatus
updates a request to the status specified, returns updated request collection with populated students
#### RequestsView.js
Contains 2 components:
- RequestsView
Parent component responsible for menu options and storing requests in 3 arrays
-- unconfirmed, confirmed, completed
- RequestCards
Takes an array of requests and the active menu item name as a prop, in addition to an update handler, time offset, and alumni id
displays requests and available actions

Screen Shot 2020-06-09 at 1 16 15 PM
Screen Shot 2020-06-09 at 1 14 53 PM
Screen Shot 2020-06-09 at 12 06 32 AM
Screen Shot 2020-06-09 at 1 15 49 PM

### DIFF
--- a/client/src/components/AlumniDirectory.js
+++ b/client/src/components/AlumniDirectory.js
@@ -79,9 +79,13 @@ export default class AlumniDirectory extends Component {
 
     async componentWillMount() {
         let result = await this.getEntries()
+        let totalPages = 0;
+        if (result.alumni !== null) {
+            totalPages = Math.ceil(result.alumni.length/pageSize);
+        }
         this.setState({
             entries: result.alumni,
-            totalPages: Math.ceil(result.alumni.length/pageSize),
+            totalPages: totalPages,
             numEntries: result.alumni.length
         })
         this.populateSearchDropdownStates(this.state.entries)
--- a/client/src/components/RequestsView.js
+++ b/client/src/components/RequestsView.js
@@ -79,10 +79,9 @@ export default class RequestsView extends Component {
     }
 
     populateConfirmedTimes(requests) {
-        let confirmedTimes = requests.map(confirmedRequest => {
+       return requests.map(confirmedRequest => {
             return (confirmedRequest.time[0].id)
         })
-        return confirmedTimes
     }
 
     handleMenuClick = (e, { id }) => this.setState({ activeItem: id })
@@ -181,11 +180,10 @@ class RequestCards extends Component {
     // This allows the component to update its state should a prop value change
     async componentWillReceiveProps({requests}) {
         await this.setState({requests: requests})
-        this.constructDisplay(this.props.requests)
+        this.constructDisplay(this.state.requests)
     }
     // This ensures that the component doesn't use an old prop on menu change
-    async componentWillMount() {
-        await this.setState({requests: this.props.requests})
+    componentWillMount() {
         this.constructDisplay(this.props.requests)
     }
 
--- a/client/src/components/RequestsView.js
+++ b/client/src/components/RequestsView.js
@@ -278,7 +278,7 @@ class RequestCards extends Component {
         if (!requests || requests.error) {
             swal({
                 title: "Error!",
-                text: "There was an error this request, please try again.",
+                text: "There was an error updating this request, please try again.",
                 icon: "error",
             });
         } else {
--- a/client/src/components/RequestsView.js
+++ b/client/src/components/RequestsView.js
@@ -1,21 +1,61 @@
 import React, { Component } from 'react';
-import { Route, BrowserRouter as Router, Switch, Redirect } from 'react-router-dom'
-import { Menu, Label } from 'semantic-ui-react';
-import Navbar from './Navbar'
+import { Menu, Label, Card, Grid, Image, Button, Segment } from 'semantic-ui-react';
 import { makeCall } from '../apis'
 
+export const timeSlotOptions = [
+    '12am - 1am',
+    '1am - 2am',
+    '2am - 3am',
+    '3am - 4am',
+    '4am - 5am',
+    '5am - 6am',
+    '6am - 7am',
+    '7am - 8am',
+    '8am - 9am',
+    '9am - 10am',
+    '10am - 11am',
+    '11am - 12pm',
+    '12p - 1pm',
+    '1pm - 2pm',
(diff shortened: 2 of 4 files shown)


server/routes/requests.js
```diff
@@ -74,4 +74,68 @@ router.post('/addRequest', async (req, res, next) => {
...
+    let requests = []
+    try {
+        let request = await requestSchema.findById(requestId);
+        let matching_queries = await requestSchema.find({
+            'status': 'Confirmed',
+            'time.id': 'request.id'
+        })
+        if (await (matching_queries === [])) {
+            console.log('Success')
+        }
```
awilli70, line 92: I'm a little stuck here - I don't want an alumni to be able to confirm a meeting when they already have one scheduled for the same time - at present I have a warning on the home page, but I'd rather it not let them hit approve. NOTE: Might be easier to accomplish on frontend

awilli70: Filtering now performed on front end

client/src/components/AlumniDirectory.js
```diff
@@ -79,9 +79,13 @@ export default class AlumniDirectory extends Component {
...
         let result = await this.getEntries()
+        let totalPages = 0;
+        if (result.alumni !== null) {
+            totalPages = Math.ceil(result.alumni.length/pageSize);
+        }
         this.setState({
             entries: result.alumni,
-            totalPages: Math.ceil(result.alumni.length/pageSize),
+            totalPages: totalPages,
             numEntries: result.alumni.length
```
awilli70, line 89: This fixed a weird bug. I'm still not entirely sure what caused it, but if there were no alumni it would throw an error

server/routes/requests.js
```diff
@@ -74,4 +74,61 @@ router.post('/addRequest', async (req, res, next) => {
     }
 });
 
+router.patch('/updateStatus/:id/:timeOffset', async (req,res, next) => {
```
awilli70, line 77: Note: this will have to be renamed when I have scheduling set up, because mentees will be able to update status by canceling an appointment

client/src/components/AlumniDirectory.js
reshadbinharun: An alternative way to do this in-line would be using javascript's short-circuit evaluation like so`result.alumni && Math.ceil(result.alumni.length/pageSize)`

server/routes/requests.js
```diff
@@ -74,4 +74,61 @@ router.post('/addRequest', async (req, res, next) => {
...
+            for (let request of dbData) {
+                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
+                if (request.requesterRole === 'STUDENT') {
+                    request.requesterObj = await studentSchema.findOne({_id: request.requester})
+                } else {
+                    request.requesterObj = await alumniSchema.findOne({_id: request.requester})
+                }
+            }
+            requests.push(dbData)
+        }
```
awilli70, line 126: This is the current pattern for populating and returning student/alumni profiles - it separates them into bins related to their condition which allows for simple display on the frontend

For schedulings, this will be *much* simpler, as only the mentor field needs to be populated

client/src/components/RequestsView.js
```diff
@@ -0,0 +1,309 @@
...
+    '6pm - 7pm',
+    '7pm - 8pm',
+    '8pm - 9pm',
+    '9pm - 10pm',
+    '10pm - 11pm',
+    '11pm - 12am'
+]
+
+/*
+ * DETAILS:
```
reshadbinharun, line 34: I like this paradigm, we should start adding a `DETAILS` documentation bit for our components.

client/src/components/RequestsView.js
```diff
@@ -0,0 +1,309 @@
...
+            confirmedTimes: confirmedTimes
+        })
+    }
+
+    getRequests(timeOffset) {
+        return makeCall({}, '/request/getRequests/'+this.props.userDetails._id+'/'+timeOffset, 'get')
+    }
+
+    populateConfirmedTimes(requests) {
+        let confirmedTimes = requests.map(confirmedRequest => {
```
reshadbinharun, line 82: nit: `return requests.map...` (confirmedTimes may not be needed)

client/src/components/RequestsView.js
```diff
@@ -0,0 +1,309 @@
...
+ * confirmedTimes - array of meeting times that have already been confirmed
+ *                  used to disable approval, preventing double booking
+ */
+class RequestCards extends Component {
+    state={
+        requests: [],
+        display: []
+    }
+    // This allows the component to update its state should a prop value change
+    async componentWillReceiveProps({requests}) {
```
reshadbinharun, line 182: There it is! 😀

client/src/components/RequestsView.js
```diff
@@ -0,0 +1,309 @@
...
+        await this.setState({requests: requests})
```
reshadbinharun, line 183: Is there a reason we're using `await` here (and in componentWillMount's setState)?

client/src/components/RequestsView.js
```diff
@@ -0,0 +1,309 @@
...
+            )
+        } else if (this.props.activeSet === 'confirmed') {
+            return(
+                <Button.Group>
+                    <Button 
+                        color='blue' 
+                        as='a'
+                        href={request.zoomLink} 
+                        target="_blank"
+                        rel="noopener noreferrer"
```
reshadbinharun, line 256: Nice! I did not know about this security measure!

client/src/components/RequestsView.js
awilli70: In my experience, code relying on a state would not work as intended if it was set earlier in a synchronous function - that said, it looks like RequestCards only uses requests as a prop so it should be safe to remove the state and the set state, as well as the async

awilli70: Changed - in testing, componentWillMount can be synchronous and has been updated to reflect that
However, componentWillReceiveProps does need to be asynchronous and set state with an await. I have changed it to construct the display using the state to reflect this

PR #33, opened 2020-06-07
Alumni and Students can upload images from their profile view

reshadbinharun:
image

* Configured test s3 bucket to store alumni/student profile images (public access, you need AWS key/secret on .env to use AWS S3 api)

* Both Alumni and Student profile allow uploading of new image via modal


client/src/components/ImageUpdateModal.js
```diff
@@ -0,0 +1,115 @@
...
+
+    submit(e) {
+        e.preventDefault()
+        this.setState({
+            submitting: true
+        }, async () => {
+            try {
+                let data = new FormData();
+                data.append('imageFile', this.state.imageFile);
+                fetch(`${BACKEND}/image/${this.props.isAlumni?`alumni` : `student`}/${this.props.id}`, {
```
reshadbinharun, line 54: I didn't use the `makeCall` helper we have set up since we do not want to add the `application/json` Content-Type

PR #37, opened 2020-06-11
Fullstack/timezonedropdown

awilli70:
Note: uses a handful of acronyms pulled from https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations
A different labeling system would probably be better - lmk if you have any thoughts

New route in index.js
PATCH changeTimeZone
body: profile id, profile role, new timezone
updates timezone 

Screen Shot 2020-06-11 at 10 07 18 PM

Screen Shot 2020-06-10 at 10 27 04 PM

Also includes a lot of small changes to formatting - definitely not final, but now the website won't break on mobile devices


client/src/components/RequestModal.js
```diff
@@ -56,25 +57,29 @@ export default class RequestModal extends Component {
         }
         this.handleValueChange = this.handleValueChange.bind(this)
         this.submitRequest = this.submitRequest.bind(this)
+        this.handleOffsetChange = this.handleOffsetChange.bind(this)
+    }
+    
+    async handleOffsetChange(offset) {
+        if (this.state.timeOffset !== offset) {
+            window.location.reload()
```
awilli70, line 65: This is gross - hopefully there is a cleaner way of doing this, but using refresh profile wasn't quick enough to fetch the profile in time for alumni, which caused an error

client/src/components/RequestModal.js
```diff
@@ -56,25 +57,29 @@ export default class RequestModal extends Component {
...
     }
 
     async createAvailabilityOptions(availabilities) {
-        /* 
-         * Offset is in minutes and in the opposite direction 
-         * that you'd expect. For UTC -4 (EST), it returns 240 
-         * (not -240 to reflect being 4 hours behind)
-         * That's why there's a tiny bit of math in the payload, to conform
-         * with our date model
-         */
```
awilli70, line 74: No longer dealing with JS timeOffsets!

client/src/components/RequestModal.js
```diff
@@ -171,15 +176,24 @@ export default class RequestModal extends Component {
...
                                     src={this.state.alumni.imageURL}
                                     rounded
                                 />
+                                <TimeZoneDropdown
+                                    userDetails={this.props.userDetails}
+                                    role={this.props.role}
+                                    liftTimezone={this.handleOffsetChange}
+                                    compact={true}
+                                />
+                                <Segment compact inverted color='blue tertiary'>
```
awilli70, line 194: Made a warning here because it doesn't feel like expected behavior - one option is to refresh at toggleModal in alumniDirectory which I think might be a better option

client/src/components/RequestsView.js
```diff
@@ -48,6 +49,22 @@ export default class RequestsView extends Component {
             confirmedTimes: []
         }
         this.handleStatusUpdate = this.handleStatusUpdate.bind(this)
+        this.handleOffsetChange = this.handleOffsetChange.bind(this)
+    }
+
+    async handleOffsetChange(offset) {
+        if (this.state.timeOffset !== offset) {
+            window.location.reload()
```
awilli70, line 57: weirdly I'm ok with refreshing here - timezone shouldn't need to be changed that frequently, so a onetime refresh shouldn't be problematic

The only reason I don't like refresh in the requestModal is that it pulls you out of the modal

client/src/components/TimeZoneDropdown.js
```diff
@@ -0,0 +1,169 @@
+import React, { Component } from 'react';
+import { Dropdown } from 'semantic-ui-react'
+import { makeCall } from '../apis'
+
+var timeZoneOptions = [
```
awilli70, line 5: Here are the tz options - I think it needs reworking but I'm not sure how to approach it

server/routes/index.js
```diff
@@ -135,6 +135,29 @@ router.get('/isLoggedIn', passport.authenticate('jwt', {session: false}), (req,
   res.json({message: "You have a fresh cookie!"});
 });
 
+router.patch('/changeTimeZone/', async (req, res, next) => {
```
awilli70, line 138: I put the route in index - I think it's too niche to go in it's own route file, and index typically deals with profile wide changes

awilli70:
Moving the dropdown to the navbar - will be doing a lot of reworking

client/src/components/Navbar.js
```diff
@@ -38,11 +40,25 @@ export default class Navbar extends Component {
             )
         })
     }
+
+    handleOffsetChange() {
+        window.location.reload()
```
awilli70, line 45: I would have liked to avoid refresh, but for one reason or another it would simply not return the updated userDetails prop in time to the component - this meant that if a user changed their timezone, it wouldn't always update the component they were viewing

Since this will usually be a onetime fire and forget feature, where it hopefully only needs to be set once, I don't think a reload is a big problem

client/src/components/Navbar.js
```diff
@@ -38,11 +40,25 @@ export default class Navbar extends Component {
...
     render() {
-        const { navItems, activeItem } = this.props
+        const { navItems, activeItem, role } = this.props
         return (
-            <Menu>
+            <Menu stackable>
                 {this.renderMenuItems(navItems, activeItem)}
+                
+                {this.props.timezoneActive &&
+                    <TimeZoneDropdown
```
awilli70, line 55: I put the timezone dropdown in the navbar - this way they should be able to notice pretty quickly if their timezone is wrong

client/src/App.js
```diff
@@ -213,9 +213,10 @@ class App extends Component {
     }
   }
 
-  async refreshProfile(role, id) {
```
reshadbinharun, line 216: If you merge this first, I will have to modify calls to refreshProfile in AlumniProfile. This change makes sense though!

client/src/components/Navbar.js
reshadbinharun: Agreed. I'm on board.

client/src/components/RequestModal.js
reshadbinharun: Good riddance!

client/src/components/TimeZoneDropdown.js
reshadbinharun: What reworkings did you have in mind?

client/src/components/TimeZoneDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+                role: this.props.userRole,
+                timeZone: parseInt(value)
+            }, '/changeTimeZone', 'patch')
+        } catch (e) {
+            console.log('timezone update failed')
+        }
+    }
+
+    handleDropdownChange = (e, { name, value }) => {
+        this.setState({ [name]: value })
```
reshadbinharun, line 155: Can you try 
```
this.setState({ ... }, () => {
this.updateTimeZone(value);
this.props.liftTimezone(value);
})
```
and see if you have any luck with timezone updating as expected? Then again you aren't using `this.state.value` as args for the methods that execute after the async setState

client/src/components/Navbar.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { Menu } from 'semantic-ui-react'
 import { Link } from 'react-router-dom'
+import  TimeZoneDropdown  from './TimeZoneDropdown'
 
 /*
```
reshadbinharun, line 6: Let's update documentation for props that are allowable here!

client/src/components/TimeZoneDropdown.js
awilli70: The issue with timezone updating is actually from refreshProfile and the way that components receive updated props - just like I had to use componentWillReceiveProps in schedulings and requests, I would need something similar in each component that receives userDetails.
ComponentDidUpdate would be a decent candidate for this, but seems to run as frequently as it possibly can which is bad for performance. componentWillReceiveProps works pretty well for schedulings and requests but seemed to be unreliable for the dropdown possibly because it requires a complete re-render

client/src/App.js
awilli70: I can actually change this back - I was attempting to use refreshProfile with NavBar but ended up not using it.

PR #40, opened 2020-06-15
Request Modal updates

awilli70:
#### Request Modal:
In createAvailabilityOptions, fetches the confirmed times to prevent students from requesting an unavailable time slot
Minor QoL fix: Availabilities are now sorted Monday-Sunday, 12am to 11pm

#### requests.js
New route:
`GET` /request/getConfirmed/:id/:timeOffset
returns a list of confirmed requests after applying the correct timezone

### DIFF
--- a/client/src/components/RequestModal.js
+++ b/client/src/components/RequestModal.js
@@ -93,7 +93,7 @@ export default class RequestModal extends Component {
             }
         }
         availabilityOptions = await availabilityOptions.sort((a, b) => {
-            let valueA = (dayOptions.indexOf(a.day) * 10000) + a.time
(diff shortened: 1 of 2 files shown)


client/src/components/RequestModal.js
```diff
@@ -67,20 +68,35 @@ export default class RequestModal extends Component {
...
+                availabilityOptions.push({
+                    day: option.day,
+                    time: option.time,
+                    key: option.id,
+                    text: readableOption,
+                    value: option.id,
+                })
+            }
         }
+        availabilityOptions = await availabilityOptions.sort((a, b) => {
```
reshadbinharun, line 95: Do we need the `await` here?

PR #44, opened 2020-06-18
Newsfeed - initial creation and new alumni

awilli70:
Created newsSchema
created newsfeed route, accessed through /news/

created `GET` /news/getNews/:role - gets all news items relevant to a specific role
created `GET` /util/allNews/ - returns all news regardless of role
created `GET` /util/data/clear/news/ - deletes all news records
added news deletion to /util/data/clear/all/

added a news item when an alumni registers an account (in /alumni/)

Created NewsFeed component, added to homepage

NewsFeed:
image

New Student:
image

newsSchema object
image


client/src/components/AlumniDirectory.js
```diff
@@ -192,7 +192,7 @@ export default class AlumniDirectory extends Component {
         let display = [];
         let i = 0;
         for (let post of this.state.entries) {
-            if (this.state.filter !== 'all') {
+            if (this.state.filter !== 'all' && post[this.state.filter]) {
```
awilli70, line 195: fixes search behavior until full indexable support added

server/routes/alumni.js
```diff
@@ -149,6 +150,11 @@ router.post('/', async (req, res, next) => {
         );
         await alumni_instance.save();
         await user_instance.save();
+        const news_instance = new newsSchema({
+            event: 'New Alumni',
+            alumni: [alumni_instance._id],
+        })
+        await news_instance.save();
```
awilli70, line 157: Template for adding a news item - most fields have defaults or aren't required, so the only necessary field is the event.
This means we can relatively cleanly make news items for multi-alumni/student activities!

client/src/components/NewsFeed.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    constructDisplay() {
+        let display = []
+        let i = 0
+        for (let feedItem of this.state.newsItems) {
+            if (feedItem.event === 'New Alumni') {
+                display.push(<Divider key={i}/>)
+                display.push(this.createNewAlumniPost(feedItem))
+            }
```
awilli70, line 40: For each new event type, we just need to add a conditional here

client/src/components/NewsFeed.js
```diff
@@ -0,0 +1,105 @@
...
+    async componentWillMount() {
+        let newsItems = await this.getNews(this.props.userRole)
+        await this.setState({
+            newsItems: newsItems.news.reverse()
+        })
+        this.constructDisplay()
+    }
+
+    getNews(role) {
+        return makeCall({}, '/news/getNews/'+{role}, 'get')
```
reshadbinharun, line 30: Don't need the curly braces around `role`

client/src/components/NewsFeed.js
reshadbinharun: Nice! Since there are going to be several types of newsItems, let's set this up as a `switch/case`

client/src/components/NewsFeed.js
```diff
@@ -0,0 +1,105 @@
...
+    /*
+     * EVENT: 'New Alumni'
+     * Display to: 'BOTH'
+     * Contains: Alumni[0] and time (stored as string) ONLY
+     * On Click: Opens a profile modal that can not be interacted with
+     */
+    createNewAlumniPost(feedItem) {
+        let alumniDetails = feedItem.alumni[0];
+        let elapsedTime = new Date();
+        let uploadedTime = new Date(feedItem.time);
```
reshadbinharun, line 55: Let's make a TODO to move this calculation to the backend, at the time we use end-point to get feedItems.
We can use [momentjs](https://stackoverflow.com/questions/42186717/moment-js-show-diff-between-2-dates)

server/routes/utilMongoose.js
```diff
@@ -532,6 +535,22 @@ router.get('/allCompanies', async (req, res) => {
     }
 })
 
+/* Newsfeed Routes */
+router.get('/allNews', async (req, res) => {
+    try {
+        let news = await newsSchema.find()
```
reshadbinharun, line 541: We can do the `timeElapsed` calculation here

server/models/newsSchema.js
```diff
@@ -0,0 +1,15 @@
+var mongoose = require('mongoose')
+
+const Schema = mongoose.Schema;
+
+const newsSchema = new Schema(
+    { 
+        event: {type: String, enum: [ 'New Alumni', 'New Student', 'Confirmed Meeting' ] }, 
+        alumni: [{type: Schema.Types.ObjectId, ref: 'Alumni'}], 
+        students: [{type: Schema.Types.ObjectId, ref: 'Student'}], 
+        time: {type: Date, default: Date.now},
```
reshadbinharun, line 10: Let's call this field `dateCreated`?

server/routes/newsfeed.js
```diff
@@ -0,0 +1,22 @@
...
+router.get('/getNews/:role', async (req, res, next) => {
+    let role = req.params.role
+    try {
+        let dbData;
+        if (role == 'ALUMNI') {
+            dbData = await newsSchema.find({role: {$ne: 'STUDENT'}}).populate('alumni').populate('students')
+        } else {
+            dbData = await newsSchema.find({role: {$ne: 'ALUMNI'}}).populate('alumni').populate('students')
+        }
+        res.json({'news' : dbData});
```
reshadbinharun, line 15: We can use momentjs to calculate the timeElapsed here

reshadbinharun:
Actually, on second thought, let's add 2 more layers of personalized to the news feed:
1. the end point `getNews` should take an additional param `:id` with role. If the role is student, they should only see newsItems for their own grade
2. the :id can be used to determine the school of the student/alumni, so students and alumni only ever see news items from their own school. For the purposes of this PR, we will need to add a `schoolId` to NewsItem schema. (We would need to add an optional grade field for newsItems that are meant for STUDENT or BOTH)

server/models/newsSchema.js
```diff
@@ -0,0 +1,15 @@
...
+const Schema = mongoose.Schema;
+
+const newsSchema = new Schema(
+    { 
+        event: {type: String, enum: [ 'New Alumni', 'New Student', 'Confirmed Meeting' ] }, 
+        alumni: [{type: Schema.Types.ObjectId, ref: 'Alumni'}], 
+        students: [{type: Schema.Types.ObjectId, ref: 'Student'}], 
+        time: {type: Date, default: Date.now}, 
+        role: {type: String, default: 'BOTH', enum: ['BOTH', 'ALUMNI', 'STUDENT']} 
+    }
```
reshadbinharun, line 12: Let's add a `school` field

server/routes/alumni.js
```diff
@@ -149,6 +150,11 @@ router.post('/', async (req, res, next) => {
         );
         await alumni_instance.save();
         await user_instance.save();
+        const news_instance = new newsSchema({
+            event: 'New Alumni',
+            alumni: [alumni_instance._id],
```
reshadbinharun, line 155: we can add `school` here

server/routes/newsfeed.js
```diff
@@ -0,0 +1,22 @@
+var express = require('express');
+var router = express.Router();
+var newsSchema = require('../models/newsSchema');
+require('mongoose').Promise = global.Promise
+
+router.get('/getNews/:role', async (req, res, next) => {
```
reshadbinharun, line 6: we can have this take an additional param `:id` (student/alumni mongo id)

server/routes/newsfeed.js
```diff
@@ -0,0 +1,22 @@
...
+    let role = req.params.role
+    try {
+        let dbData;
+        if (role == 'ALUMNI') {
+            dbData = await newsSchema.find({role: {$ne: 'STUDENT'}}).populate('alumni').populate('students')
```
reshadbinharun, line 11: we can filter by school (for both student and alumni) here

awilli70:
Note: probably shouldn't merge quite yet - receiving an error upon registration
Registering an alumni creates a new account successfully, but raises a negative swal and shows an error (Unauthorized 401 from alumni add)

client/src/components/NewsFeed.js
```diff
@@ -0,0 +1,144 @@
...
+export default class NewsFeed extends Component {
+    state={
+        newsItems: [],
+        display: []
+    }
+
+    async componentWillMount() {
+        let newsItems = await this.getNews(this.props.userRole)
+        if (newsItems.news) {
+            await this.setState({
```
reshadbinharun, line 25: nit: Don't need `await` here

client/src/components/NewsFeed.js
```diff
@@ -0,0 +1,144 @@
...
+        let i = 0
+        for (let feedItem of this.state.newsItems) {
+            switch (feedItem.event) {
+                case 'New Alumni':
+                    display.push(<Divider key={i}/>)
+                    display.push(this.createNewAlumniPost(feedItem))
+                    break;
+                case 'New Student':
+                    display.push(<Divider key={i}/>)
+                    display.push(this.createNewStudentPost(feedItem))
```
reshadbinharun, line 47: Need a `break` here

server/models/newsSchema.js
```diff
@@ -0,0 +1,19 @@
...
+const newsSchema = new Schema(
+    { 
+        event: {type: String, enum: [ 'New Alumni', 'New Student', 'Confirmed Meeting' ] }, 
+        alumni: [{type: Schema.Types.ObjectId, ref: 'Alumni'}], 
+        students: [{type: Schema.Types.ObjectId, ref: 'Student'}], 
+        school: {type: Schema.Types.ObjectId, ref: 'School'},
+        dateCreated: {type: Date, default: moment()}, 
+        role: {type: String, default: 'BOTH', enum: ['BOTH', 'ALUMNI', 'STUDENT']},
+        timeElapsed: {type: String, required: false},
+        grade: {type: Number, default: null}
```
reshadbinharun, line 15: Let's make this required `false`

server/routes/newsfeed.js
```diff
@@ -0,0 +1,31 @@
...
+    let role = req.params.role
+    let id = req.params.id
+    try {
+        let dbData;
+        if (role == 'ALUMNI') {
+            userInfo = await alumniSchema.findById(id)
+            dbData = await newsSchema.find({role: {$ne: 'STUDENT'}, school: userInfo.school}).populate('alumni').populate('students')
+        } else {
+            userInfo = await studentSchema.findById(id)
+            dbData = await newsSchema.find({role: {$ne: 'ALUMNI'}, school: userInfo.school, grade: {$in: [null, userInfo.grade]}}).populate('alumni').populate('students')
```
reshadbinharun, line 19: Should we also pull newsItems that have `null` for grade when the role is `STUDENT`?

server/models/newsSchema.js
awilli70: for grade? I use null as way to store posts that don't require grade info so that I can use null as an accepted value

server/routes/newsfeed.js
awilli70: Yes - null is used for posts where grade is irrelevant (ie all both)

server/models/newsSchema.js
```diff
@@ -0,0 +1,19 @@
...
+
+const newsSchema = new Schema(
+    { 
+        event: {type: String, enum: [ 'New Alumni', 'New Student', 'Confirmed Meeting' ] }, 
+        alumni: [{type: Schema.Types.ObjectId, ref: 'Alumni'}], 
+        students: [{type: Schema.Types.ObjectId, ref: 'Student'}], 
+        school: {type: Schema.Types.ObjectId, ref: 'School'},
+        dateCreated: {type: Date, default: moment()}, 
+        role: {type: String, default: 'BOTH', enum: ['BOTH', 'ALUMNI', 'STUDENT']},
+        timeElapsed: {type: String, required: false},
```
reshadbinharun, line 14: nit: We don't need to store this, as it will only ever be calculated at display time. We can `newsItem.map (item => Object.assign({}, ...item, timeElapsed: moment(item.dateCreated).fromNow()))` when we're returning a newsItem.

server/models/newsSchema.js
reshadbinharun: Yep! Wouldn't setting it to be `required: false` also only allow us to use it when needed?

server/routes/newsfeed.js
reshadbinharun: Ah I see! Future-proofing for other posts! Makes sense.

server/models/newsSchema.js
reshadbinharun: Nvm, this works!

server/models/newsSchema.js
awilli70: It wasn't ever being stored - rather it was a workaround for the fact that each item is a specific document type and would only allow assignment to fields of the model
That said, I've removed the field, and included the map which converts everything to objects before adding timeElapsed

PR #50, opened 2020-06-20
Fullstack/request feedback

awilli70:
Added a modal to requestsview (final note) and schedulingsview (feedback)
Added a button to completed cards to trigger modals
Added endpoints to update requests with feedback/final notes
Removed mark as completed from schedulings

Request/Scheduling cards will now only show populated fields (mentee note, final note, feedback)

Images:

Schedulings completed:
image

Schedulings feedback modal:
image

Requests completed:
image

Requests final note modal:
image

### DIFF
--- a/server/routes/requests.js
+++ b/server/routes/requests.js
@@ -178,7 +178,7 @@ router.patch('/leaveFinalNote/:id/:timeOffset', passport.authenticate('jwt', {se
         for (let status of conditions) {
             const dbData = await requestSchema.find({mentor: alumniId, status: status})
             for (let request of dbData) {
-                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
+                request.time = timezoneHelpers.applyTimezone(request.time, timeOffset)
                 if (request.requesterRole === 'STUDENT') {
                     request.requesterObj = await studentSchema.findOne({_id: request.requester})
                 } else {
@@ -265,7 +265,7 @@ router.patch('/updateScheduling/:id/:role/:timeOffset', passport.authenticate('j
                     status: status
                 }).populate('mentor')
             for (let request of dbData) {
-                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
+                request.time = timezoneHelpers.applyTimezone(request.time, timeOffset)
             }
             schedulings.push(dbData)
         }
--- a/server/routes/requests.js
+++ b/server/routes/requests.js
@@ -164,6 +164,38 @@ router.patch('/updateRequest/:id/:timeOffset', passport.authenticate('jwt', {ses
     }
 })
 
+router.patch('/leaveFinalNote/:id/:timeOffset', passport.authenticate('jwt', {session: false}), async (req,res, next) => {
+    let alumniId = req.params.id;
+    let timeOffset = req.params.timeOffset;
+    let requestId = req.body.requestId;
+    let finalNote = req.body.finalNote;
+    let conditions = ['Awaiting Confirmation', 'Confirmed', 'Completed']
+    let requests = []
+    try {
+        let request = await requestSchema.findById(requestId);
+        request.finalNote = finalNote
+        await request.save()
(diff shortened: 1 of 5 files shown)


server/routes/requests.js
```diff
@@ -164,6 +164,38 @@ router.patch('/updateRequest/:id/:timeOffset', passport.authenticate('jwt', {ses
...
+    let requests = []
+    try {
+        let request = await requestSchema.findById(requestId);
+        request.finalNote = finalNote
+        await request.save()
+        for (let status of conditions) {
+            const dbData = await requestSchema.find({mentor: alumniId, status: status})
+            for (let request of dbData) {
+                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
+                if (request.requesterRole === 'STUDENT') {
```
reshadbinharun, line 182: I'm wondering if we can move this conditional to populate requestorObj in the `POST:addRequest` route, and not have to worry about populating it on any of these other end-points. I don't imagine the requesterObj to change once a request has been created, right?

server/routes/requests.js
```diff
@@ -164,6 +164,38 @@ router.patch('/updateRequest/:id/:timeOffset', passport.authenticate('jwt', {ses
...
+    let conditions = ['Awaiting Confirmation', 'Confirmed', 'Completed']
+    let requests = []
+    try {
+        let request = await requestSchema.findById(requestId);
+        request.finalNote = finalNote
+        await request.save()
+        for (let status of conditions) {
+            const dbData = await requestSchema.find({mentor: alumniId, status: status})
+            for (let request of dbData) {
+                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
```
reshadbinharun, line 181: We can remove `await` here

server/routes/requests.js
```diff
@@ -244,6 +276,37 @@ router.patch('/updateScheduling/:id/:role/:timeOffset', passport.authenticate('j
...
+        updatedScheduling.feedback = feedback
+        await updatedScheduling.save()
+        for (let status of conditions) {
+            const dbData = await requestSchema.find({
+                    requester: requesterId,
+                    requesterRole: requesterRole,
+                    status: status
+                }).populate('mentor')
+            for (let request of dbData) {
+                request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
```
reshadbinharun, line 298: We can remove `await` here!

server/routes/requests.js
awilli70: That depends more on what we do with Alumni/ Student objects
Populating on return is a mongoose paradigm designed to prevent storing the same information multiple times, which avoids the same profile being kept in 2 different states - If requester could only be a student/alumni, then I'd only be storing a ref

requesterObj is a workaround for working with and returning Mongo documents
Documents do not allow addition of fields not in the schema, but I still want to populate the requester in the same way that I would populate the mentor using
`requestSchema.find({}).populate('Alumni')`
Here mongoose just replaces the alumni field with the correct document, but since I cannot replace a string with an Object (due to mongoose's rules) the way I replicated model.populate() was to have a separate field that can hold a generic object - but that generic object (and field) will never be stored on the database (functionally, it doesn't exist as an object unless it is on the front end)

I ended up doing something similar in newsfeed.js when you asked me to replace timeElapsed - but there I used map and had to convert everything to an object before adding the field. 
I'm not sure what the overhead on .toObject() is, but I decided adding a field that isn't kept on the backend to requests wasn't super expensive and was in keeping with mongoose's design

awilli70:
Changed the color to teal to match the labels showing how many unconfirmed/confirmed schedulings/requests were present - orange/yellow were a bit too harsh

server/routes/requests.js
reshadbinharun: Seems fair! We might revisit our design for requestSchema in the future if too many requests to refetch prove to be a bottle-neck further down the road, but this shouldn't be a concern for now!

PR #57, opened 2020-06-25
Meeting Settings Tab

awilli70:
Removed the meeting related buttons from Alumni Profile
Otherwise, more or less a movement of code from one file to another

image


client/src/components/AlumniMentorship.js
```diff
@@ -88,6 +166,93 @@ export default class AlumniMentorship extends Component {
...
+            <ZoomUpdateModal
+                modalOpen={this.state.zoomUpdateOpen}
+                zoomLink={zoomLink || ''}
+                closeModal={this.closeZoomUpdateModal}
+                id={details._id}
+            />
+            </>
+        )
+
+        let topicListItems = []
```
reshadbinharun, line 238: nit: We could use [this list](https://react.semantic-ui.com/elements/list/#types-divided) to make this slightly more aesthetic

PR #58, opened 2020-06-26
Extended Feedback, Added pre-meeting note

awilli70:
### Alumni

image

image

image

### Student

image

image

image

### DIFF
--- a/client/src/components/AlumniMentorship.js
+++ b/client/src/components/AlumniMentorship.js
@@ -388,8 +388,10 @@ class RequestCards extends Component {
         requests: [],
         display: [],
         showFeedbackModal: false,
+        showAlumniNoteModal: false,
         requestDetails: null,
-        finalNote: ''
+        finalNote: '',
+        alumniNote: ''
     }
     // This allows the component to update its state should a prop value change
     async componentWillReceiveProps({requests}) {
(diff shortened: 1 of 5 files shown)


client/src/components/StudentMentorship.js
```diff
@@ -365,7 +398,7 @@ class SchedulingCards extends Component {
                         <Button onClick={this.toggleFeedbackModal.bind(this)}>
                             Done
                         </Button>
-                        <Button onClick={this.submitFinalNote.bind(this)} primary>
+                        <Button onClick={this.submitFinalNote.bind(this)} primary disabled={!this.state.publicFeedback || !this.state.privateFeedback}>
```
awilli70, line 401: Students should not be able to submit without having public or private feedback for the alumni
Testimonials are more special and doesn't have an immediate impact on alumni, and thus should only be submitted when other feedback is given

server/routes/requests.js
```diff
@@ -132,6 +132,7 @@ router.patch('/updateRequest/:id/:timeOffset', passport.authenticate('jwt', {ses
             const dbData = await requestSchema.find({mentor: alumniId, status: status}).populate('student')
             for (let request of dbData) {
                 request.time = await timezoneHelpers.applyTimezone(request.time, timeOffset)
+                request.privateFeedback = undefined;
```
awilli70, line 135: This removes any evidence of privateFeedback from returned documents

client/src/components/AlumniMentorship.js
```diff
@@ -536,6 +553,30 @@ class RequestCards extends Component {
         })
     }
 
+    toggleAlumniNoteModal(e) {
+        let requestDetails = this.props.requests.find(request => request._id === e.currentTarget.getAttribute('requestid'))
+        console.log(requestDetails)
```
reshadbinharun, line 558: Remove debug log here

client/src/components/StudentMentorship.js
```diff
@@ -167,7 +167,9 @@ class SchedulingCards extends Component {
         schedulings: [],
         display: [],
         schedulingDetails: [],
-        feedback: '',
+        publicFeedback: '',
```
reshadbinharun, line 170: very appropriate field names!

client/src/components/StudentMentorship.js
reshadbinharun: Makes sense! let's go with it

server/routes/requests.js
reshadbinharun: Perfect! And we're not losing any of this data on fetches, so that's great

PR #63, opened 2020-07-01
Image required for sign up + Alumni must provide one topic of consultancy + zoomLink

reshadbinharun:
Changes:
* Alumni and Student are both required to upload an image during registrations
* Alumni must also provide at least one topic of consultancy and a zoom meeting link

Screens:
image

image

### DIFF
--- a/server/routes/image.js
+++ b/server/routes/image.js
@@ -10,6 +10,10 @@ require('dotenv').config();
 var alumniSchema = require('../models/alumniSchema');
 var studentSchema = require('../models/studentSchema');
 require('mongoose').Promise = global.Promise
+var bcrypt = require('bcrypt');
+var moment = require('moment');
+
+const HASH_COST = 4;
 
 // configure the keys for accessing AWS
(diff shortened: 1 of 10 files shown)


server/routes/image.js
```diff
@@ -87,4 +89,30 @@ router.post('/student/:studentId', async (req, res, next) => {
...
+                    imageUrl: imageLocation
+                })
+            }
+        })
+    } catch (e) {
+        console.log("Error: image#addImage", e);
+        res.status(500).send({'error' : e});
+    }
+});
+
```
awilli70, line 117: I saw your text earlier - this solution is fine for now, especially since it works and has the behavior we want
One option we have (That can be filed under Nice to Have alongside the mongo expiry) is having a temporary bucket for images that then moves images to the standard bucket upon verification and deletes old files automatically (https://stackoverflow.com/questions/46769995/temporarily-upload-on-s3-and-remove-if-its-not-used)

PR #64, opened 2020-07-01
Added networking view, get route for conversations

awilli70:
Current View:
Screen Shot 2020-07-01 at 12 29 17 AM

### DIFF
--- a/client/src/components/AlumniNetworking.js
+++ b/client/src/components/AlumniNetworking.js
@@ -37,7 +37,6 @@ export default class AlumniNetworking extends Component {
     constructConversation(conversation) {
         let userIndex = conversation.alumni.findIndex(item => item._id === this.props.userDetails._id);
         let recipientIndex = (userIndex + 1) % 2;
-        let labelActive = conversation.messages[0].senderId[0] !== this.props.userDetails._id ? true : false;
         return(
             <Grid key={conversation._id} columns={'equal'}>
                 <Grid.Row columns={2}>
@@ -52,7 +51,7 @@ export default class AlumniNetworking extends Component {
                     <Grid.Column>
                         <Card fluid>
                             <Card.Content>
-                            {labelActive && 
+                            {!conversation.seen[userIndex] && 
                                 <Label color='teal' corner='right' icon='envelope' />
                             }
                                 <Card.Header>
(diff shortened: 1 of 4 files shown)


server/models/conversationSchema.js
```diff
@@ -16,7 +16,9 @@ const conversationSchema = new Schema(
         alumni: [{type: Schema.Types.ObjectId, ref: 'Alumni'}],
         messages: [messageSchema],
         seen: [{type: Boolean}],
-        dateCreated: {type: Date, default: moment()}, 
+        dateCreated: {type: Date, default: moment()},
+        //RETURN FIELD NOT KEPT IN DB
+        timeFromMessage: {type: String}
```
awilli70, line 21: comment is pretty self explanatory, but this field is a workaround for returning fields that don't need to be stored on the database

I looked into mongoose's virtuals, and while they have the right functionality they aren't passed in a response unless I go through and convert every document to JSON before sending the response

client/src/components/AlumniNetworking.js
```diff
@@ -0,0 +1,79 @@
+import React, { Component } from 'react';
+import { Card, Image, Grid, Label, icon } from 'semantic-ui-react'
```
reshadbinharun, line 2: `icon` import may not be needed if used as a prop on `Label`

client/src/components/AlumniNetworking.js
```diff
@@ -0,0 +1,79 @@
...
+        let conversations = await this.getConversations()
+        this.setState({
+            timeOffsetHours: timeOffsetHours,
+            conversations: conversations.conversations
+        })
+        this.createDisplay(conversations.conversations)
+    }
+
+    getConversations() {
+        return makeCall({}, '/conversations/get/' + this.props.userDetails._id, 'get')
```
reshadbinharun, line 26: `/get/` part of path may be redundant

server/routes/conversations.js
```diff
@@ -41,11 +42,29 @@ router.post('/add/', passport.authenticate('jwt', {session: false}), async (req,
...
     } catch (e) {
-        console.log('request/addConversation Error: ' + e)
+        console.log('conversations/add Error: ' + e)
         res.status(500).send({
             message: 'Failed creating conversation: ' + e
         });
     }
 });
 
+router.get('/get/:id', passport.authenticate('jwt', {session: false}), async (req, res) => {
```
reshadbinharun, line 52: `/get` may be redundant here

PR #67, opened 2020-07-08
added gradYear validation

awilli70:
Screen Shot 2020-07-08 at 12 42 03 PM

### DIFF
--- a/client/src/components/Signup.js
+++ b/client/src/components/Signup.js
@@ -57,6 +57,7 @@ export default class Signup extends React.Component {
             grade: null, // required
             // ALUMNI ONLY
             graduationYear: null, // required
+            graduationYearValid: true,
             // location
             country: '',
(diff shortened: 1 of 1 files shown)


client/src/components/Signup.js
```diff
@@ -796,9 +810,9 @@ export default class Signup extends React.Component {
                             type="email"
                             required={true}
                             style={fieldStyle}
-                            error={!this.state.emailValid}
+                            error={!this.state.emailValid && this.state.email !== ''}
                         >
-                            {!this.state.emailValid ? getErrorLabel('Please enter a valid email!') : null}
+                            {!this.state.emailValid && this.state.email !== '' ? getErrorLabel('Please enter a valid email!') : null}
```
awilli70, line 815: hides the error label when email is empty but other fields have been filled out

PR #66, opened 2020-07-07
removed email field from alumni and student objects

awilli70:
Note: This PR makes significant changes to the student and alumni schema.
After merging this PR, do the following in shell to ensure that mongoose will not throw errors:
~$ mongo
mongo shell: $ use ofi-testdata
mongo shell: $ db.dropDatabase()
mongo shell: $ exit

then, assuming you're using homebrew, restart your mongo drivers by doing:
~$ brew services stop mongodb-community
~$ brew services start mongodb-community
To ensure that the local mongodb instance fully restarts. This has the added benefit of not having to frequently restart mongodb, as it will instead start on machine startup in the future.

This will resolve any E11000 errors caused by mongoose anticipating an email field in alumni/students

This has been successfully tested on features with email integration such as requests and signup
Note: for security reasons, user routes should probably be deleted - however, this warrants further discussion

PR #68, opened 2020-07-09
admin creation and route setup

awilli70:
user.role is now an array
Users can switch between roles using a dropdown if more than one is available:  
Screen Shot 2020-07-09 at 1 17 58 PM

Currently defaults to Alumni/Student view first, though this relies upon the order of the array in mongodb

Placeholder admin routes now available
Since this represented a major change to the database, updating may cause E11000 errors from mongodb - use PR #66 for reference.

There are likely a couple of bugs introduced with this update - from what I could tell, every component was still working because they receive role as a prop, but since role is now an array there may be a couple of problems in higher level components

### DIFF
--- a/client/src/components/Header.js
+++ b/client/src/components/Header.js
@@ -241,7 +241,8 @@ export default class Header extends Component {
     }
 
     renderLogo() {
-        let imageLink = this.props.loggedIn ? this.props.schoolLogo : require('./logo.png')
+        let imageLink = this.props.loggedIn && this.state.currRole !== 'ADMIN' ? 
+            this.props.schoolLogo : require('./logo.png');
         return (
                 <Image centered src={imageLink} size='small'/>                
         )
--- a/client/src/components/Header.js
+++ b/client/src/components/Header.js
@@ -255,6 +255,7 @@ export default class Header extends Component {
             <>
                 <Label basic >Current Role:
                     <Dropdown
+                        compact
                         style={{'marginLeft': '5px'}}
                         selection
                         options={this.state.availableRoles}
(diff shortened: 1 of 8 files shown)


client/src/components/Header.js
```diff
@@ -228,11 +241,60 @@ export default class Header extends Component {
...
+                </Label>
+                
+            </>
+            }
+            </div>
+        )
+    }
+
+    changeRole = (e, { value }) => {
+        this.setState({ currRole: value })
```
reshadbinharun, line 273: We might want to do `this.setState({ currRole: value} ,() => this.props.liftRole(value))`

client/src/components/Header.js
```diff
@@ -228,11 +241,60 @@ export default class Header extends Component {
...
+        this.props.liftRole(value)
+    }
+
+    fetchUser() {
```
reshadbinharun, line 277: async

server/routes/index.js
```diff
@@ -53,8 +53,10 @@ router.post('/login', (req, res, next) => {
             return next(error);
           }
           try {
-            let userRole = user.role && user.role.toUpperCase()
-            if (userRole === "ALUMNI") {
+            let userRole = user.role && user.role.map(role => {
+              return role.toUpperCase()
+            })
+            if (userRole[0] === "ALUMNI") {
```
reshadbinharun, line 59: Can we do `userRole.includes("ALUMNI")`?

server/routes/index.js
```diff
@@ -67,7 +69,7 @@ router.post('/login', (req, res, next) => {
                   details: alumni
                 }
               );
-            } else if (userRole === "STUDENT") {
+            } else if (userRole[0] === "STUDENT") {
```
reshadbinharun, line 72: Can we do `userRole.includes("STUDENT")`?

PR #74, opened 2020-07-20
SchoolList setup

awilli70:
Screen Shot 2020-07-20 at 3 11 32 PM

TODO: Create Add school Modal

Note: In the process of writing this module, I once again ran into the image upload bug - my theory (though it may not be a good one) is that multiparty is not a good form parser. While the makeCallWithImage API is successful and contains an image, the parser does not get the file quickly enough.

Image Select Modal works because its route has to hash an email before accessing the file, which adds enough delay to find the image and upload it.

New school and modal:

Screen Shot 2020-07-22 at 9 11 35 PM

Screen Shot 2020-07-22 at 9 05 53 PM

### DIFF
--- a/server/routes/admin.js
+++ b/server/routes/admin.js
@@ -177,7 +177,7 @@ router.patch('/mergeColleges/:adminid', passport.authenticate('jwt', {session: f
     let adminId = req.params.adminId
     let colleges = req.body.items
     let name = req.body.name
-    let location = await collegeSchema.findById(colleges[0])
+    let school = new schoolSchema
     let country = location.country
     try {
         if (!isAdmin(adminId)) {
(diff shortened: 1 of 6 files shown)


reshadbinharun:
I think we should be good to merge this, with a //TODO in the makeCallWithImage API since that is a known bug.

server/routes/admin.js
```diff
@@ -185,4 +201,26 @@ router.patch('/mergeColleges/:adminid', passport.authenticate('jwt', {session: f
...
+        if (!isAdmin(adminId)) {
+            res.status(400).send('Invalid Admin ID');
+            return;
+        }
+        let school = new schoolSchema({
+            name: name,
+            country: country
+        })
+        await school.save();
+        res.status(200).send({'message': 'Successfully merged colleges'})
```
reshadbinharun, line 219: revise message!

PR #76, opened 2020-07-22
Allow AWS Deployment

reshadbinharun:
This PR allows the root-level npm start script to serve build react app from index router


server/app.js
```diff
@@ -33,14 +33,14 @@ app.use(express.json());
...
-app.use(express.static(path.join(__dirname, 'public')));
+app.use(express.static(path.join(__dirname, 'build')));
 
 /* 
  * If testDB is true, uses a locally hosted mongoDB
  * Otherwise, it will use a cloud hosted DB set in the .env file
  * MongoDB must be installed
  */
-const testDB = true;
+const testDB = process.env.DEV_MODE;
```
awilli70, line 43: To prevent truthy string weirdness, should probably be testDB = (process.env.DEV_MODE.toLowerCase() === “true”);

PR #77, opened 2020-07-27
College Rep portal

awilli70:
Screen Shot 2020-07-27 at 7 16 13 PM

Created College Rep Schema, with same fields as admin schema + college reference

To add a college rep to the app, send a GET request to the server at /util/newCollegeRep

Also includes a hotfix for admin, where alumni can now switch to the admin dashboard from any alumni tab (rather than just the newsfeed)

NOTE: Current base is development, not develop - After review I will switch base to develop and merge, and delete development

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -285,14 +285,19 @@ class App extends Component {
 
   async fetchProfile(role, id) {
     let result;
-    if (role === 'STUDENT') {
-      result = await makeCall({}, ('/student/one/'+id), 'get')
-    } else if (role === 'ALUMNI') {
-      result = await makeCall({}, ('/alumni/one/'+id), 'get')
-    } else if (role === 'ADMIN') {
-      result = await makeCall({}, ('/admin/one/'+id), 'get')
-    } else if (role === 'COLLEGE_REP') {
-      result = await makeCall({}, ('/collegeRep/one/'+id), 'get')
+    switch (role) {
+      case STUDENT:
+        result = await makeCall({}, ('/student/one/'+id), 'get');
+        break;
+      case ALUMNI:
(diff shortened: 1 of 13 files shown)


client/src/App.js
```diff
@@ -268,6 +291,8 @@ class App extends Component {
       result = await makeCall({}, ('/alumni/one/'+id), 'get')
     } else if (role === 'ADMIN') {
```
reshadbinharun, line 292: nit: can we have this method be `switch/case`?

client/src/App.js
```diff
@@ -295,7 +320,13 @@ class App extends Component {
 
   liftRole(role) {
     this.setState({
-      role: role
+      role: role,
+      roleChanged: true
+    }, () => {
+      console.log(this.state.roleChanged)
```
reshadbinharun, line 326: Remove log here

awilli70:
> Looks great! For the hotfix, I was wondering if it might make more sense to 1. pass a handler method down to dropdown that changes roles, and the handler can change a state `resetToRootRoute` in `App` that can cause the redirect? Let me know what you think!

I might be misinterpreting your comment, but I think(?) it does that already, using similar logic to the loginForm  

The handler method (liftRole) is passed to the dropdown via the header, which switches the role in App to match the dropdown.  

In this handler it also sets roleChanged to true, with a callback to immediately set it back to false (in essence resetToRootRoute). Since roleChanged is referenced in the render function, any changes to its state will trigger a render which means it will redirect to / when set to true and then immediately be set back to false.

The difference between LoginForm and redirecting in App.js is that the router is defined in the render method of App (which consequently means that any redirects must also live there) whereas LoginForm lives as a subcomponent and can return a redirect upon successful login

PR #82, opened 2020-07-30
College shortlist

irenelam22:
Features:
- Displays the user's college shortlist
- Disables submit upon duplicates


client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,93 @@
...
+    closeCollegeModal() {
+        this.setState({
+            collegeModalOpen: false
+        }, () => {
+            this.props.refreshProfile(STUDENT, this.props.details._id)
+        })
+    }
+
+    findDuplicate = (newCollege) => {
+        return this.props.details.colleges.some(college => college == newCollege);
```
reshadbinharun, line 34: Let's do a `===` comparison (type and value)

server/models/studentSchema.js
```diff
@@ -15,7 +15,8 @@ const studentSchema = new Schema(
     school: {type: Schema.Types.ObjectId, ref: 'School', required: true},
     schoolLogo: {type: String, default:'https://cdn.pixabay.com/photo/2015/10/05/22/37/blank-profile-picture-973460_1280.png'},
     isModerator: {type: Boolean, default: false},
-    interests: {type: Array, required: false}
+    interests: {type: Array, required: false},
+    colleges: [{type: Schema.Types.ObjectId, ref: 'College'}]
```
reshadbinharun, line 19: Can we name this field `collegeShortlist`?

server/routes/students.js
```diff
@@ -33,6 +34,29 @@ router.post('/', async (req, res, next) => {
         const emailVerified = false
         const approved = false
         const verificationToken = crypto({length: 16});
+
+        // find or create College
```
reshadbinharun, line 38: We probably don't need to change this post end-point as students would not add colleges to their shortlist when they sign up

reshadbinharun:
This looks like it's off to a solid start!
I may have misled you a little in inviting you to edit the `CollegeUpdateModal` to handle adding a single college to a shortlist. The `CollegeUpdateModal` is meant for alumni, and it allows alumni to add a college -- it may not be worthwhile trying to refactor this modal to handle adding a single college to a shortlist!
I think what might make more sense is creating a new `AddToShortlistModal` that uses a `PooledSingleSelectDropdown` with allowAddition={false}. This would ensure that the student is only able to select a college from a list of available colleges, and not be able to add a college.
I think we would have to worry about students adding a duplicate college if we go with this approach.

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                </Menu.Item>
+                
+            </Menu>
+            {this.state.activeItem === 'shortlist' &&
+                <div style={{paddingLeft: 13, paddingRight: 13}}>
+                    <Card fluid>
+                    <Card.Content>
+                    <Card.Header>{details.name || 'Unavailable'}</Card.Header>
+                    <Card.Description>Colleges: {details.colleges.map(e => e.name).join(', ') || 'Unavailable'}
```
reshadbinharun, line 62: May be using [this component](https://react.semantic-ui.com/elements/list/#types-divided) might serve the aesthetic you were looking for?

server/routes/students.js
```diff
@@ -151,4 +175,37 @@ router.patch('/interests/add/:id', async (req, res, next) => {
     }
 })
 
+router.patch('/college/update/:id', passport.authenticate('jwt', {session: false}), async (req, res, next) => {
```
reshadbinharun, line 178: If we do allow a single college addition, we can just send the id of the college and push into the `collegeShortlist` field in a student record! We wouldn't need to include the logic to check if the college was a new one vs. an existing one!

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,103 @@
...
+        })
+    }
+    closeCollegeModal() {
+        this.setState({
+            collegeModalOpen: false
+        }, () => {
+            this.props.refreshProfile(STUDENT, this.props.details._id)
+        })
+    }
+    shortlist(update) {
```
reshadbinharun, line 32: This method could be called `addToShortlist` and it can update the `collegeShortlist` state variable here like ...

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,103 @@
...
+        alert('We pass argument from Child to Parent: ' + update);
+        this.setState({
+            collegeShortlist: update,
```
reshadbinharun, line 35: let currentCollegeShortlist = this.state.collegeShortlist
currentCollegeShortlist.push(newCollege) // `update` in your case
this.setState({
  shortlist: currentCollegeShortlist
})

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,103 @@
...
+    constructor(props) {
+        super(props)
+        this.state = {
+            activeItem: "shortlist",
+            collegeModalOpen: false,
+            collegeShortlist: [],
+        }
+        this.openCollegeModal = this.openCollegeModal.bind(this);
+        this.closeCollegeModal = this.closeCollegeModal.bind(this);
+        var shortlist = this.shortlist.bind(this);
```
reshadbinharun, line 17: this would then need to be bound like so ...
```
this.addToShortlist = this.addToShortlist.bind(this);
```

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,103 @@
...
+                                    size="mini"
+                                    onClick={this.openCollegeModal}
+                                >
+                                    <Icon name='pencil' color="blue"/>
+                                </Button>
+                                <CollegeShortlistModal
+                                    modalOpen={this.state.collegeModalOpen}
+                                    id={details._id}
+                                    collegeShortlist={details.collegeShortlist}
+                                    shortlist = {shortlist.bind(this)}
```
reshadbinharun, line 88: you can simply pass in `addToShortlist` as `addToShortlist={this.addToShortlist}`

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,206 @@
...
+                        goodInput: true
+                    }, () => {
+                        swal({
+                            title: "Done!",
+                            text: "Your college has been successfully updated!",
+                            icon: "success",
+                        })
+                    })
+                    console.log(this.props.collegeShortlist);
+                    this.props.collegeShortlist.push(this.state.existingCollegeName);
```
reshadbinharun, line 128: this line could then be replaced by `this.props.addToShortlist(findDuplicate)`

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,206 @@
...
+
+    clearState() {
+        this.setState({
+            country: '',
+            existingCollegeId: '',
+            existingCollegeName: ''
+        })
+    }
+
+    findDuplicate = (newCollege) => {
```
reshadbinharun, line 76: nit: since this returns a boolean can we name the method `isDuplicateCollegeEntry`

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,206 @@
...
+        }, async () => {
+            let payload = {
+                collegeCountry: this.state.country,
+                existingCollegeId: this.state.existingCollegeId,
+                existingCollegeName: this.state.existingCollegeName
+            }
+            try {
+                const result = await makeCall(payload, `/student/college/update/${this.props.id}`, 'patch')
+                if (this.findDuplicate(this.state.existingCollegeName)) {
+                    this.setState({
```
reshadbinharun, line 94: nit: the error modal works but may be we can disable the submit button is the user enters a duplicate college...

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,206 @@
...
+                            value={this.state.country}
+                            onChange={this.handleCountrySelection}
+                        />
+                    }
+                </Modal.Content>
+                <Modal.Actions>
+                    <Button 
+                        // disabled={!this.state.newCollege && !this.state.existingCollegeId}
+                        onClick={
+                            this.submit
```
reshadbinharun, line 189: ... like so `disabled={this.isDuplicateCollegeEntry}`

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,110 @@
...
+    openCollegeModal() {
+        this.setState({
+            collegeModalOpen: true
+        })
+    }
+    closeCollegeModal() {
+        this.setState({
+            collegeModalOpen: false
+        }, () => {
+            this.props.refreshProfile(STUDENT, this.props.details._id)
```
awilli70, line 29: refreshProfile will refetch the student prop but it won't actually update the state of a component - You probably want a componentWillReceiveProps(nextprops) lifecycle method, which will call addToShortlist

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+    addToShortlist(newCollege) {        
+        let currentCollegeShortlist = this.props.details.collegeShortlist
+        currentCollegeShortlist.push(newCollege) 
+        this.setState({
+            shortlist: currentCollegeShortlist
+        })
+        // let currentCollegeShortlist = this.state.collegeShortlist
+        this.state.collegeShortlist.push(newCollege);
+    }
```
awilli70, line 40: If you're able to get this working with a update lifecycle method, remove this code

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,110 @@
...
+
+    render() {
+        const details = this.props.details;
+        const isViewOnly = this.props.isViewOnly;
+        // if (details.collegeShortlist.size > 0) {
+            
+        //     this.collegeShortlist = details.collegeShortlist.map(e => e.name).join(', ');
+        // }
```
awilli70, line 48: Remove commented out code in future PRs

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,110 @@
...
+                <Menu.Item
+                    id='shortlist'
+                    name='College Shortlist'
+                    active={this.state.activeItem === 'shortlist'}
+                    onClick={this.handleMenuClick}
+                >
+                    College Shortlist
+                    {   (this.state.shortlist !== []) && 
+                         <Label color='teal'>{details.collegeShortlist.length}</Label>
+                        //  <Label color='teal'>{this.state.collegeShortlist.length}</Label>
```
awilli70, line 62: Same Here

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,110 @@
...
+                </Menu.Item>
+                
+            </Menu>
+            {this.state.activeItem === 'shortlist' &&
+                <div style={{paddingLeft: 13, paddingRight: 13}}>
+                    <Card fluid>
+                    <Card.Content>
+                    <Card.Header>{details.name || 'Unavailable'}</Card.Header>
+                    {/* <Card.Description>Colleges: {this.state.collegeShortlist[0] || 'Unavailable'} */}
+                    <Card.Description>Colleges: {details.collegeShortlist.map(e =>
```
awilli70, line 73: This is probably the reason your state related version didn't work - You're constructing the shortlist based on the props only, when you are generally making changes to the state to try and trigger updates

awilli70: It might be worth only working with state for this Component
You might need to add a component will / did mount to set the initial state, but it will make other changes easier and will be simpler to reflect, because if a state is referenced in render method and it changes, the component will re-render automatically

server/routes/students.js
```diff
@@ -151,4 +153,24 @@ router.patch('/interests/add/:id', async (req, res, next) => {
...
+    try {
+        // find or create College
+        // const collegeCountry = req.body.collegeCountry
+        let existingCollegeId = req.body.existingCollegeId
+        var college
+        if (existingCollegeId) {
+            college = await collegeSchema.findOne({_id: existingCollegeId})
+        }
+        let student = await studentSchema.findOne({_id: req.params.id})
+        student.collegeShortlist.push(college._id);
```
awilli70, line 166: This seems like unnecessary duplication - why check for an existing college, fetch the existing college, and then push the id of the fetched college to the shortlist when you have the id already?

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,191 @@
...
+                            search
+                            selection
+                            options={this.state.countryOptions}
+                            value={this.state.country}
+                            onChange={this.handleCountrySelection}
+                        />
+                    }
+                </Modal.Content>
+                <Modal.Actions>
+                    <Button
```
awilli70, line 175: This button need to be primary

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,100 @@
...
+                
+            </Menu>
+            {this.state.activeItem === 'shortlist' &&
+                <div style={{paddingLeft: 13, paddingRight: 13}}>
+                    <Card fluid>
+                    <Card.Content>
+                    <Card.Header>{details.name || 'Unavailable'}</Card.Header>
+                    <Card.Description>Colleges: {details.collegeShortlist.map(e => 
+                        <li>{e.name}</li> || <li>{e}</li>) || 'Unavailable' }
+                        {
```
awilli70, line 69: Move button in line with colleges using Semantic UI grid

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,100 @@
...
+        }, () => {
+            this.props.refreshProfile(STUDENT, this.props.details._id)
+        })
+    }
+
+    addToShortlist(newCollege) {        
+        let currentCollegeShortlist = this.props.details.collegeShortlist
+        currentCollegeShortlist.push(newCollege) 
+        this.setState({
+            shortlist: currentCollegeShortlist
```
reshadbinharun, line 37: I think we want
```
this.setState({
collegeShortlist: currentCollegeShortlist
})
```
...

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,100 @@
...
+        })
+        this.state.collegeShortlist.push(newCollege);
```
reshadbinharun, line 39: ... and won't need this ([directly changing `this.state`](https://daveceddia.com/why-not-modify-react-state-directly/#:~:text=Mutating%20state%20directly%20can%20lead,its%20parent%20re%2Drenders).) isn't advisable)

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,187 @@
...
+                    })
+                    this.props.closeModal();
+                } 
+                else {
+                    this.setState({
+                        submitting: false,
+                    }, () => {
+                        swal({
+                            title: "Done!",
+                            text: "Your college has been successfully updated!",
```
reshadbinharun, line 110: nit: college shortlist*

client/src/components/CollegeShortlist.js
awilli70: I think this code is no longer used by anything and can be removed (the shortlist is constructed purely using the props)

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,99 @@
...
+                         <Label color='teal'>{details.collegeShortlist.length}</Label>
+                    }
+                </Menu.Item>
+                
+            </Menu>
+            {this.state.activeItem === 'shortlist' &&
+                <div style={{paddingLeft: 13, paddingRight: 13}}>
+                    <Card fluid>
+                    <Card.Content>
+                    <Card.Header>{details.name || 'Unavailable'}</Card.Header>
```
reshadbinharun, line 65: I think this can be { `${details.name}'s College Picks` }

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,99 @@
...
+                    <Card.Description>Colleges: {details.collegeShortlist.map(e =>
```
reshadbinharun, line 66: One suggestion here would be:
```
{details.collegeShortlist.length ? 
<Card.Description> ... </Card.Description>
: 
<Card.Description> Please add some colleges by country </Card.Description>
}
``` 
So that we gracefully handle an empty state

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,99 @@
...
+                </Menu.Item>
+                
+            </Menu>
+            {this.state.activeItem === 'shortlist' &&
+                <div style={{paddingLeft: 13, paddingRight: 13}}>
+                    <Card fluid>
+                    <Card.Content>
+                    <Card.Header>{details.name || 'Unavailable'}</Card.Header>
+                    <Card.Description>Colleges: {details.collegeShortlist.map(e => 
+                        <li key={e.id}>{e.name}</li> || <li>{e}</li>) || 'Unavailable' }
```
reshadbinharun, line 67: I think you can remove ` || <li>{e}</li>) || 'Unavailable' ` here

client/src/components/CollegeShortlist.js
```diff
@@ -0,0 +1,99 @@
...
+                            !isViewOnly ? 
+                            <>
+                                <Button
+                                    style={{'margin': '0 0 2px 2px'}}
+                                    icon
+                                    inverted
+                                    size="mini"
+                                    onClick={this.openCollegeModal}
+                                >
+                                    <Icon name='pencil' color="blue"/>
```
reshadbinharun, line 78: Let's Actually have a button here with text `Add College`

client/src/components/CollegeShortlistModal.js
reshadbinharun: bump on this, it should actually be primary!

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,188 @@
...
+                            selection
+                            options={this.state.countryOptions}
+                            value={this.state.country}
+                            onChange={this.handleCountrySelection}
+                        />
+                    }
+                </Modal.Content>
+                <Modal.Actions>
+                    <Button 
+                        variant="primary"
```
reshadbinharun, line 172: I think using `primary` instead of `variant="primary"` might do! (styles currently not applying)

server/routes/students.js
```diff
@@ -151,4 +153,20 @@ router.patch('/interests/add/:id', async (req, res, next) => {
     }
 })
 
+router.patch('/college/update/:id', passport.authenticate('jwt', {session: false}), async (req, res, next) => {
```
reshadbinharun, line 156: could this end-point start with `/collegeShortlist/update/:id`

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,188 @@
+import React, { Component } from 'react';
+import {Button, Modal, Grid, Label, Icon, Dropdown } from 'semantic-ui-react';
+import PooledSingleSelectDropdown from "./PooledSingleSelectDropdown"
+import { makeCall } from "../apis";
+import swal from "sweetalert";
+
+/*
+    Modal makes a PATCH call to update alumni information
```
reshadbinharun, line 8: nit: Update description of modal (and the props it receives)

client/src/components/CollegeShortlistModal.js
```diff
@@ -0,0 +1,188 @@
...
+                            options={this.state.countryOptions}
+                            value={this.state.country}
+                            onChange={this.handleCountrySelection}
+                        />
+                    }
+                </Modal.Content>
+                <Modal.Actions>
+                    <Button 
+                        primary
+                        disabled={this.isDuplicateCollegeEntry(this.state.existingCollegeName)}
```
reshadbinharun, line 173: nit: Can be handled in a follow-up story, but the button should also be disabled (and `loading={this.state.submitting}`) when `this.state.submitting` is true

PR #89, opened 2020-08-19
hotfix for logout, made button layout mobile friendly

awilli70:
Mobile:
Screen Shot 2020-08-19 at 3 39 38 PM

Desktop:
Screen Shot 2020-08-19 at 3 39 55 PM

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -225,7 +225,6 @@ class App extends Component {
       approved: false,
       role: null,
       roleChanged: false,
-      schoolId: '',
       userDetails: {}
     };
     this.logout = this.logout.bind(this);
@@ -311,7 +310,9 @@ class App extends Component {
   async logout() {
     await makeCall({}, '/logout', 'get');
     this.setState({
-      loggedIn: false
+      loggedIn: false,
+      userDetails: {},
+      role: null
     });
   }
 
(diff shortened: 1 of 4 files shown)


client/src/App.js
```diff
@@ -311,7 +310,9 @@ class App extends Component {
   async logout() {
     await makeCall({}, '/logout', 'get');
     this.setState({
-      loggedIn: false
+      loggedIn: false,
+      userDetails: {},
```
reshadbinharun, line 314: phewwww!

client/src/App.js
```diff
@@ -692,17 +693,7 @@ class App extends Component {
         )
       default:
         return (
-          <Route exact path={`/`} render={(props) =>
```
reshadbinharun, line 695: What this causing the fleeting error on login earlier?

awilli70: If role was being set to null, then yes! The purpose of this was to make it a little smoother when logging in (it would flash a placeholder screen before it properly fetched the profile data) - but this meant if you logged off from anywhere that wasn't news, it would say page not found (/profile doesn't exist in the default route case)
Because of changes to the initial data fetch, and other changes to make the current header possible, the default route didn't need to exist and could just redirect to /login

PR #97, opened 2020-09-04
College Shortlist - Deletes colleges

irenelam22:
Warning: Will not work if duplicate colleges already exist in the list (assumes that 'Add college' properly handles duplicates)


reshadbinharun:
Archiving for use as code not mergeable in current state.

PR #103, opened 2020-09-19
Submit Bug

miriam-shamash:
Fixed submit button to bar user from submitting multiple times while the submit processes.

### DIFF
--- a/client/src/components/AlumniContactModal.js
+++ b/client/src/components/AlumniContactModal.js
@@ -122,7 +122,7 @@ export default class AlumniContactModal extends Component {
                     <Button
                         primary
                         onClick={this.submitRequest}
-                        disabled={this.state.message === ''}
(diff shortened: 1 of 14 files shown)


reshadbinharun (changes requested):
Great work! Let's also disable Submit buttons on the `CollegeShortlistModal`, `ImageUpdateModal`, `InterestsUpdateModal`, `TopicPreferencesModal`, and `ZoomUpdateModal`.
On the buttons let's also add the `loading={this.state.submitting}` to enabled the spinner on the buttons as we submit!
Lastly, as housekeeping let's add an screenshot for before vs. after for one submitting state!

reshadbinharun (approved):
Looks great! Once you are able to locally verify the disabled and loading states (by slowing down the network tab) on these interactions, feel free to go ahead and merge!

PR #104, opened 2020-09-26
Image File size Check on Upload/Select

kiranmisner:
Should work now for files less than 5MB are the only ones allowed to upload.


client/src/components/ImageSelectModal.js
```diff
@@ -23,12 +23,21 @@ export default class ImageSelectModal extends Component {
...
         this.fileTypeIsImage = this.fileTypeIsImage.bind(this)
+        this.checkFileSizeLimit = this.checkFileSizeLimit.bind(this)
     }
 
     fileTypeIsImage (file) {
         return file && (file.type === 'image/jpeg' || file.type === 'image/jpg' || file.type === 'image/png')
     }
 
+    checkFileSizeLimit(file) {
+        if (file.size > 5 * 1024 * 1024) {
```
reshadbinharun, line 34: Let's have a top level constant FILE_SIZE_LIMIT const at top of file.
Also since we're return a boolean, we could simply `return file.size > FILE_SIZE_LIMIT`

client/src/components/ImageSelectModal.js
```diff
@@ -88,8 +109,8 @@ export default class ImageSelectModal extends Component {
                         primary
                         onClick={this.submit}
                         loading={this.state.loading}
-                        disabled={!this.state.imageFile || !this.fileTypeIsImage(this.state.imageFile) || this.state.loading}
-                    >
+                        disabled={!this.state.imageFile || !this.fileTypeIsImage(this.state.imageFile) || (this.state.fileSize === false)}
```
reshadbinharun, line 112: This ought to be `imageSize` right?

client/src/components/ImageSelectModal.js
```diff
@@ -43,6 +52,18 @@ export default class ImageSelectModal extends Component {
                 icon: "warning",
             });
         }
+        if (this.checkFileSizeLimit(file)) {
+            this.setState({
+                imageSize: true
```
reshadbinharun, line 57: We could not introduce a new variable here, and simply set imageFile to null here

client/src/components/ImageUpdateModal.js
```diff
@@ -15,20 +15,31 @@ export default class ImageUpdateModal extends Component {
...
         this.fileTypeIsImage = this.fileTypeIsImage.bind(this)
+        this.checkFileSizeLimit = this.checkFileSizeLimit.bind(this)
     }
 
     fileTypeIsImage (file) {
         return file && (file.type === 'image/jpeg' || file.type === 'image/jpg' || file.type === 'image/png')
     }
 
+    checkFileSizeLimit(file) {
+        if (file.size > 5 * 1024 * 1024) {
```
reshadbinharun, line 32: same comment as above!

client/src/components/ImageUpdateModal.js
```diff
@@ -15,20 +15,31 @@ export default class ImageUpdateModal extends Component {
...
+            return false; 
+        } else {
+            return true;
+        }
+    }
+
     selectFile(e) {
         e.preventDefault();
         let file = e.target.files[0]
+        this.checkFileSizeLimit(file);
```
reshadbinharun, line 42: This might be left here in error

reshadbinharun (changes requested):
Good job putting the business logic together! Left some comments.
Also let's cherry-pick the changes for code related to this work only, let me know if you need help with that!

reshadbinharun:
Already implemented!

PR #108, opened 2020-10-04
Ensure user has scrolled to bottom of terms and conditions page (Might not get merged)

reshadbinharun:
This PR seems a little Redundant as the user will need to get to bottom of scrollable div to dismiss it


reshadbinharun:
PR is redundant since we use a checkbox to affirm that contract has been read

PR #121, opened 2020-11-15
Weekly digest email subscription

reshadbinharun:
* Changes:
- `user` record now has `emailSubscribed` and `emailSubscriptionToken`. `emailSubscribed` is originally set to true for both alumni and students. Every email for weekly digest will contain a link to unsubscribe that will toggle the `emailSubscribed` flag to false, and reset the `emailSubscriptionToken`

- a cron job is added that will run once every Sunday, which will find all conversations that took place between alumni, for each alumni, and send the summary to each alumni

- email template in emailbodyBuilder is cleaned up to make it more readable

* Manual testing steps:
1. Comment out send `await sg.send(emailObject, true)` in sendAlumnusEmailDigest and replace with a console.log statement with contents of `emailObject` [Do do not commit this change]
2. Change cron expression to trigger more frequently for testing [Do not commit]. Also change the line in getWeeklyMessagesSummaryForAlumni to match with cronExpression
3. Deploy code to test environment
4. Create test 3 test alumni gmail accounts, and sign up on test environment
5. As alumni 1 send messages to alumni 2 and 3, wait for cron expression worth of time, and check for logs. Alumni 2 and 3 should receive digest containing message counts from alumni 1. Wait another cron expression worth of time, and confirm the same notifications are not received.


craragon77 (approved):
looks great! thanks for doing this!

PR #151, opened 2021-01-21
Introduce logging library

reshadbinharun:
Added winston as logging library, introduced it in the student router only to start off.


miriam-shamash (approved):

server/routes/students.js
```diff
@@ -99,7 +101,7 @@ router.post('/', async (req, res, next) => {
             student: student_instance
         });
     } catch (e) {
-        console.error("Error: students#one", e)
+        logger.error(`POST | action=student/ | email=${email} | error=${e}`)
```
sbetti02, line 104: Looks to me like we might want a function to format the logs in this format you want? Maybe it could just take the request type/route from the request, the error, and optionally any additional metadata as an object?

sbetti02 (approved):
Looks good to me, just a suggestion on how we could generalize these log points to make it more flexible/easier to work with in the future

PR #159, opened 2021-02-04
Hotfix: Toggling between filters should clear searchText

reshadbinharun:


client/src/components/AlumniDirectory.js
```diff
@@ -455,7 +460,7 @@ export default class AlumniDirectory extends Component {
             case 'gradYear':
                 return (
                     <Dropdown 
-                        placeholder='Year:'
+                        placeholder={'Graduation Year'}
```
sbetti02, line 463: Do the placeholders that are just a string need to be encapsulated as JSX? I think it should be fine to just leave it as `placeholder="Graduation Year"` and `placeholder="Interest"`?

server/routes/index.js
```diff
@@ -329,6 +329,30 @@ router.get('/totalCounts', async (req, res, next) => {
   }
 })
 
+router.get('/sampleSignUps', async (req, res, next) => {
```
sbetti02, line 332: Can you add a line describing what this route is for?

client/src/landingPageContent/JoinUs.js
```diff
@@ -20,10 +22,63 @@ export default function JoinUs (props) {
...
+            <Segment>
+                <Header as='h4'>Get a chance to meet mentors like...</Header>
+                <Grid
+                    stackable
+                    textAlign='center'
+                >
+                    {
+                        sampleSignUps.map(alumnus => {
+                            let transitionDuration = [1000, 2000, 3000][Math.floor(Math.random() * [1000, 2000, 3000].length)]
+                            return (
```
sbetti02, line 50: It feels like these cards could deserve their own component

sbetti02 (commented):
Looks good to me, just a couple minor comments

client/src/landingPageContent/JoinUs.js
reshadbinharun: I don't think these cards will get reused anywhere else so I don't think that's worth the effort.

server/routes/index.js
reshadbinharun: addressed

PR #162, opened 2021-02-12
Ambassador portal

reshadbinharun:
Admins are able to promote alumni to become ambassadors of their respective countries (country determined by their high school context). Ambassadors can control privileges of alumni, students of their country, onboard new schools for their country, and send out polls and announcements to schools for their country.

Bonus: Welcome badge is now a thing!

Screens:
image
Welcome badge
image
User able to toggle between roles
image
Among privileges for Ambassadors, they can grant access levels upto interschool
image
Admins can now promote alumni to country ambassadors


client/src/App.js
```diff
@@ -809,6 +865,86 @@ class App extends Component {
           />
         </>
         )
+      case COUNTRY_AMBASSADOR:
+        return (
```
sbetti02, line 869: Personally, I find this whole chunk a bit difficult to read.. Wondering if there's a way to generalize these cases to simplify this code?

client/src/components/Header.js
```diff
@@ -275,16 +280,70 @@ export default class HeaderComponent extends Component {
...
+                return (
+                    <Label
+                        style={{
+                            backgroundColor: black,
+                            color: 'white'
+                        }}
+                    >
+                        Welcome {this.props.name.split(' ')[0]} (Alumnus)
+                    </Label>
+                )
```
sbetti02, line 307: This cases are all the same except for the name that it's given.. Seems like this could be its own small component where you pass in just the name you're trying to welcome and the role.
Looking at it more closely, if you pass the name + currentRole into this component, you can make this whole thing here just one line

client/src/App.js
sbetti02: Yeah similar to the comment I made further down, I kinda wonder if you could pass a small set of props to a subcomponent and generalize this code in a really reusable way

client/src/components/admin_dashboard/SchoolsList.js
```diff
@@ -18,10 +26,23 @@ export default function SchoolsList(props) {
 
     const pageSize = 4;
 
+    const urlBuilder = (path) => {
```
sbetti02, line 29: Do we need 3 different urlBuilder functions? I kinda feel like these could all be unified and put into a util file

server/routes/country_ambassador.js
```diff
@@ -0,0 +1,448 @@
...
+    return alumni
+}
+
+/*
+    Return all students with the school and access context information populated
+    The country is determined from location of school
+    @param country
+    @return array of student objects
+*/
+async function fetchAllStudentsWithAccessContexts(country) {
```
sbetti02, line 47: This and the above function can be unified, the only difference I can see is whether you use the studentSchema or the alumniSchema

server/routes/country_ambassador.js
```diff
@@ -0,0 +1,448 @@
...
+        if (type === 'ALUMNI') {
+            let alumni = await alumniSchema.findById(profileId);
+            newApprovalState = !alumni.approved
+            alumni.approved = newApprovalState;
+            let userRecordForAlumnus = await userSchema.findById(alumni.user, {email: 1})
+            email = userRecordForAlumnus.email
+            name = alumni.name
+            await alumni.save()
+            dbData = await fetchAllAlumniWithAccessContexts(country)
+        } else if (type === 'STUDENT') {
```
sbetti02, line 133: These two cases can be put into a small function and unified

server/routes/country_ambassador.js
```diff
@@ -0,0 +1,448 @@
...
+            studentUsers = await studentSchema.find().where('country').in(countriesTargetted).exec()
+            alumniUsers = await alumniSchema.find().where('country').in(countriesTargetted).exec()
+        } else if (rolesTargetted.includes('ALUMNI')) {
+            alumniUsers = await alumniSchema.find().where('country').in(countriesTargetted).exec()
+        } else if (rolesTargetted.includes('STUDENTS')) {
+            studentUsers = await studentSchema.find().where('country').in(countriesTargetted).exec()
+        }
+    }
+    // by school
+    else if (schoolsTargetted.length) {
```
sbetti02, line 335: Looks like these two cases can be combined with just passing in either 'school' or 'country' into a function as the argument

sbetti02 (changes requested):
I didn't have time now to look in detail through everything, but overall, functionally this definitely looks good, but the biggest thing i'm seeing is places where code could be combined to make it more readable and maintainable

client/src/App.js
reshadbinharun: Hmm, definitely agree. The App.js component is unwieldily clunky, I'll take a bit to extract this piece out

client/src/components/Header.js
reshadbinharun: Nice catch! Changing

reshadbinharun: Actually, I think there's among the 3 roles, with the string interpolation and Flag component, there are enough differences to warrant this explicit breakdown, what do you think?

server/routes/country_ambassador.js
reshadbinharun: noted, makes sense!

server/routes/country_ambassador.js
reshadbinharun: noted, will change!

server/routes/country_ambassador.js
reshadbinharun: I think since we're just ever going to target by school or country, we'd be losing readability trading in maintainability like that. I don't think we're gaining too much abstracting away school vs country here, happy to have more back and forth on this!

server/routes/country_ambassador.js
reshadbinharun: addressed!

server/routes/country_ambassador.js
reshadbinharun: addressed, thanks for the suggestion!

client/src/components/Header.js
sbetti02: For the string interpolation, I think it's all the same except for wanting to put "Ambassador" in front of the country ambassador's name? You could check for that case pretty easily since you could pass in the role to the new component.

Also yeah, similar thing for the flag component, you could kinda do the same thing as above, but I'm wondering if there's a better way.. Is props.school.country defined for other types? If not, you could just pass that in and only show the flag if you passed it a country, or something.. Feel like this is kinda hard to work through over github, so happy to chat about it if you wanna discuss more

server/routes/country_ambassador.js
sbetti02: Hm, I'm not sure if I fully understand the context for what you're talking about here, but I guess it seems like those cases, even if they don't have 100% overlap, would have a lot of the same functionality? I guess if that ends up being the case, we could still use a function that does the part for the unified section where they're both the same, but then they could each have code that extends what they do beyond the shared part

PR #164, opened 2021-02-14
Add footy points to the front end

sbetti02:
Just added the the footy points to the front end in a few locations. Also added some simple tests where possible. The one notable exception is that I didn't add  a test to the alumni directory because I think it requires more upfront cost to get that set up.


client/src/components/Header.js
```diff
@@ -5,6 +5,8 @@ import { Link } from "react-router-dom"
 import 'semantic-ui-css/semantic.min.css';
 import swal from "sweetalert";
 import { makeCall } from "../apis";
+import FootyPoints from './FootyPoints';
+
```
reshadbinharun, line 9: Get outta here with that python bullsh*t

client/src/components/Header.js
```diff
@@ -247,6 +249,7 @@ export default class HeaderComponent extends Component {
                 
             </>
             }
+            <FootyPoints points={this.props.footyPoints}/>
```
reshadbinharun: I'm thinking whether we merge the ambassador portal in first and have the FootyPoints badge live in the welcomeBadge component?  Otherwise we have 3 components in the first column of the header which I can get with. Your call!

reshadbinharun: image
Let's add some styling and be deliberate with stacking behavior.
Let's add `style = {{ margin: 5px }}` or better yet let's always get these component to stack vertically. Use a `Grid.Row`for accessContext dropdown and the badge

reshadbinharun (changes requested):
Very clean and thank you for breaking ground with f/e tests!
Let's add some ✨  styling ✨  before we merge. Also refer to comment for welcome badge and let me know what you'd like to do!

client/src/components/FootyPoints.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Label } from 'semantic-ui-react';
+const logo = require('./logo.png');
+
+function FootyPoints (props) {
+    return (
+        <Label as='a' image>
```
reshadbinharun: Why are we rending Label as an anchor tag?

client/src/components/Header.js
sbetti02: Lol I think there's actually some minor performance improvements for using import over require, and I think it's expected to be more of the standard in JS, though it's a newer feature

client/src/components/Header.js
sbetti02: Hm interesting, and yeah I'm indifferent on which to merge in first, sounds like it should get moved to the welcomeBadge component, but I think it should just be a one line thing to get shifted over? And the supporting test should still work, so doesn't seem like a big deal to me either way

sbetti02: Also yeah I'll look into the styling, good call. I was doing it without that dropdown present so it wasn't quite so obvious that it was out of line

client/src/components/FootyPoints.js
sbetti02: Bc my brain is small and I just copied + pasted the code from the example

client/src/components/Header.js
sbetti02: Also fuck you I like python :)

reshadbinharun: Nothing wrong with using `import` that's a welcome ES6 thing, but I mean the 2 line spaces. It's like you don't even try my dude

focitti (approved):

client/src/components/AlumniDirectory.js
```diff
@@ -243,6 +244,8 @@ export default class AlumniDirectory extends Component {
                             <Card.Description>Company: {post.companyName || 'Unavailable'}</Card.Description>
                             {post.interests && post.interests.length ? <Card.Description>Interests: {this.getInterests(post.interests)}</Card.Description> : null}
                             {post.topics && post.topics.length ? <Card.Description>Topics of Consultancy: {this.getTopics(post.topics)}</Card.Description> : null}
+                            <FootyPoints points={ post.footyPoints }/>
```
reshadbinharun: Can we have this be in line with the alumnus's name?

client/src/components/AlumniProfile.js
```diff
@@ -428,6 +429,12 @@ export default class AlumniProfile extends Component {
                             </Segment>
                         </Card.Content>
                 }
+                {
+                    isViewOnly ? null :
```
reshadbinharun: Same here, inline with Alumnus's name instead of in a different section?

client/src/components/WelcomeBadge.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return (
+        role ? 
+            <div>
+                <Label style={{
+                    backgroundColor: 'black',
+                    color: 'white'
+                }}>
+                    {welcomeMessage}
+                    {showFlag ?
```
reshadbinharun: The formatting here is off can we fix this?
image

So it has flag first and appropriate padding like so
image

reshadbinharun (changes requested):
Love the refactors! This PR is looking great, small suggestions.
Can we also attach screenshots to PR for f/e changes?

sbetti02:
@reshadbinharun think all the style stuff should be updated as you were saying. I found it to be more trouble than it was worth to fit the footypoints badge on the same line as the name due to it being difficult to line up the text height and the icon height, so I put it on the next line. Screenshots attached, take a look and lmk what you think
Screen Shot 2021-03-03 at 11 25 35 PM
Screen Shot 2021-03-03 at 11 26 02 PM
Screen Shot 2021-03-03 at 11 26 11 PM
Screen Shot 2021-03-03 at 11 26 23 PM
Screen Shot 2021-03-03 at 11 26 57 PM

reshadbinharun (approved):
Beautiful! Let's lift offfff

PR #169, opened 2021-03-02
Backend/add methods to allow alumni event creation

### LINKED ISSUE #176: Virtual Events backend

focitti:
This is an initial PR to set up data base & add a helper endpoint to allow alumni virtual event creation. As it stands, not entirely testable on it's own.

### DIFF
--- a/server/models/eventSchema.js
+++ b/server/models/eventSchema.js
@@ -4,13 +4,14 @@ const Schema = mongoose.Schema;
 
 const eventSchema = new Schema(
     {
-      	creator: {type: Schema.Types.ObjectId, ref: 'User', required: true},
-        name: {type: String, required: true},
-        date: {type: Date, rrequired: true},
+        creator: {type: Schema.Types.ObjectId, ref: 'User', required: true},
+        title: {type: String, required: true},
+        date: {type: Date, required: true},
         link: {type: String, required: true},
         description: {type: String, required: true},
         school: {type: Schema.Types.ObjectId, ref: 'School', required: false},
-        years: [{type: Number, required: false}]
+        startYear: {type: Number, required: false},
+        endYear: {type: Number, required: false}
     }
 );
 
--- a/server/models/eventSchema.js
+++ b/server/models/eventSchema.js
@@ -1,0 +1,17 @@
+var mongoose = require('mongoose')
+
+const Schema = mongoose.Schema;
+
+const eventSchema = new Schema(
+    {
+      	creator: {type: Schema.Types.ObjectId, ref: 'User', required: true},
+        name: {type: String, required: true},
+        date: {type: Date, rrequired: true},
+        link: {type: String, required: true},
+        description: {type: String, required: true},
+        school: {type: Schema.Types.ObjectId, ref: 'School', required: false},
+        years: [{type: Number, required: false}]
+    }
+);
+
+module.exports = mongoose.model('Event', eventSchema);
--- a/server/routes/events.js
+++ b/server/routes/events.js
@@ -1,0 +1,98 @@
+var express = require('express')
+var passport = require("passport")
+var router = express.Router()
+var schoolSchema = require('../models/schoolSchema')
+var alumniSchema = require('../models/alumniSchema')
+var userSchema = require('../models/userSchema')
+var eventSchema = require('../models/eventSchema')
+require('mongoose').Promise = global.Promise
+
+
+/*
+    Helper function to ensure user has access to a given school
+    @param accessContexts - list of access levels for the given user
+    @param userSchool - user school object
+    @param school - object of school user is attempting to access
+    @return Boolean determining if the user has access to school
+*/
+const isValidAccessContext = ({ accessContexts, userSchool, school }) => {
+    if (accessContexts.includes("GLOBAL")) {
+        return true
+    } else if (accessContexts.includes("INTERSCHOOL")) {
+        return userSchool.country === school.country
+    } else {
+        return userSchool._id.toString() === school._id.toString()
+    }
+}
+
+
+/*
+    API used by alumni to create alumni accessible events
+    @return the user created event
+*/
+router.post('/create/:userId', async (req, res, next) => {
+    try {
+        let userRecord = await userSchema.findById(req.params.userId)
+        let isAlumni = userRecord.role.includes("ALUMNI")
+        // event creation restricted to alumni only
+        if (!isAlumni) {
+            res.status(403).json({
+                message: "User must be an alumnus to create event"
+            })
+            return
+        }
+        var alumnus = await alumniSchema.findOne({ user: userRecord._id })
+        // school validation
+        var school = null
+        var userSchoolId = alumnus.school
+        var userSchool = await schoolSchema.findById(userSchoolId)
+        if (req.body.school) {
+            school = await schoolSchema.findById(req.body.school)
+            if (!school) {
+                res.status(404).json({
+                    message: `Could not find school with id ${school.id}`
+                })
+                return
+            } else if (!isValidAccessContext({ accessContexts: userRecord.accessContexts, userSchool, school })) {
+                res.status(403).json({
+                    message: `User does not have access level to create event in ${school.country}`
+                })
+                return
+            }            
+        }
+        if (!(school || userSchool)) {
+            console.error("Error: events#create. Invalid school saved on user object.")
+            res.status(404).json({
+                message: `Could not find school with id ${userSchoolId}`
+            })
+            return
+        }
+        school = school || userSchool
+        // event body validation
+        if (!req.body.name || !req.body.date || !req.body.description) {
+            res.status(400).json({
+                message: "Event must contain have name, description and date fields provided"
+            })
+            return
+        }
+        // event creation
+        var event = new eventSchema({
+            creator: userRecord._id,
+            name: req.body.name,
+            date: req.body.date,
+            link: req.body.link,
+            description: req.body.description,
+            school: school._id,
+            years: req.body.years
+        })
+        await event.save()
+
+        res.status(200).json({ event })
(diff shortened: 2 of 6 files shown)


server/routes/utilMongoose.js
```diff
@@ -97,9 +97,9 @@ const createAlumni = async (_email, _name, _country, _city, _profession, _compan
             country: _country,
             city: _city,
             jobTitle: _profession,
```
focitti, line 108: `jobTitle` and `company` are optional fields on the alumni schema and removing the struct requirement makes setting up test data a wee bit less annoying.

server/routes/utilMongoose.js
```diff
@@ -323,53 +323,28 @@ router.post('/addAlumni/', async (req, res, next) => {
...
         const location = req.body.location;
         const profession = req.body.profession;
-        const company = req.body.company;
-        const college = req.body.college;
+        const country = req.body.country;
+        const city = req.body.city;
         const availabilities = [];
-        const timeZone = req.body.timeZone;
         const zoomLink = req.body.zoomLink;
         const password = req.body.password;
```
focitti, line 376: instead of re-creating logic to add user and alumni objects, utilise `createAlumni` helper method already in use by `/util/seed`.

Did a general search and appears this method is unused so no updates needed elsewhere.

server/routes/utilMongoose.js
```diff
@@ -499,6 +474,17 @@ router.get('/data/clear/user', async (req, res, next) => {
     }
 });
 
+router.put('/updateUser/:userId', async (req, res, next) => {
```
focitti, line 523: added endpoint to crudely update user object for testing purposes.

server/tests/events.test.js
```diff
@@ -0,0 +1,162 @@
...
+    await mongoose.connection.close()
+    delete process.env.NODE_ENV
+})
+
+test('students not allowed to create new events', async () => {
+    var student = await studentSchema.findOne()
+    var userID = student.user
+    const res = await request(app).post(`/events/create/${userID}`)
+        .send({
+            eventName: 'how dinosaurs helped establish modern architecture'
```
reshadbinharun: how do I RSVP to this

server/routes/events.js
```diff
@@ -0,0 +1,98 @@
...
+        return userSchool._id.toString() === school._id.toString()
+    }
+}
+
+
+/*
+    API used by alumni to create alumni accessible events
+    @return the user created event
+*/
+router.post('/create/:userId', async (req, res, next) => {
```
reshadbinharun: We should guard this API by `passport.authenticate`

reshadbinharun (approved):
Looks great! One comment on guarding the event creation API

reshadbinharun (changes requested):
Meant to Request changes with passport.authenticate! Otherwise looks fab

server/models/eventSchema.js
```diff
@@ -0,0 +1,17 @@
+var mongoose = require('mongoose')
+
+const Schema = mongoose.Schema;
+
+const eventSchema = new Schema(
+    {
+      	creator: {type: Schema.Types.ObjectId, ref: 'User', required: true},
+        name: {type: String, required: true},
```
sbetti02: What do you think about naming this 'title' instead of 'name'? Maybe doesn't really matter, but when I was glancing over it I thought it was referring to the name of the creator for a sec

server/routes/events.js
```diff
@@ -0,0 +1,98 @@
...
+        let userRecord = await userSchema.findById(req.params.userId)
+        let isAlumni = userRecord.role.includes("ALUMNI")
+        // event creation restricted to alumni only
+        if (!isAlumni) {
+            res.status(403).json({
+                message: "User must be an alumnus to create event"
+            })
+            return
+        }
+        var alumnus = await alumniSchema.findOne({ user: userRecord._id })
```
sbetti02: I think for scoping concerns we might want to use let here instead of var for these variables

server/routes/events.js
```diff
@@ -0,0 +1,98 @@
...
+        // school validation
+        var school = null
```
sbetti02: Could we rename this to maybe something like inputSchool or something? I just feel like it gets a little confusing between this variable and the userSchool variable on lines like `if (!(school || userSchool))`

server/tests/events.test.js
```diff
@@ -0,0 +1,162 @@
...
+        // create alumnus
+        await request(app).post('/util/addAlumni/').send({
+            email: "schoolalum@ofi.com",
+            name: "school alum",
+            accessContexts: [ "INTRASCHOOL" ],
+            gradYear: Math.random(),
+            country: "Germany",
+            city: "Osgiliath"
+        })
+        var alumnus = await alumniSchema.findOne({ name: "school alum", city: "Osgiliath" })
```
sbetti02: I like what you did here, wondering if maybe this could be available as part of the initial seed data I'm imagining a specific alum/school that we can access with certain details would probably be helpful for a lot of different tests

sbetti02 (approved):
Various comments and thoughts, but overall this looks awesome! The only comment I think I'd like to see ideally would be making the distinction clearer between userSchool/school in some of the functions, but I think this is good enough to merge as is regardless, so I'm approving now

server/models/eventSchema.js
focitti: Oh facts. I like that name. Will update.

server/routes/events.js
focitti: Gooooooood catch. I forgot to up-date that post adding tests.

server/routes/events.js
focitti: Sure I can update this. I usually use var and assume things are well defined, but that is a good point.

server/tests/events.test.js
focitti: In the MVP version RSVP'ing won't be a part of the functionality. There is a follow up story to address this though. Mayhaps this is something we can talk about during stand up as well.

server/routes/events.js
sbetti02: Oh one thought, I don't think we should submit the userId with the request, I think we should be able to just figure that out from the request details of the logged in user? Noticed this while building the front end part

server/models/eventSchema.js
```diff
@@ -0,0 +1,17 @@
...
+        date: {type: Date, rrequired: true},
```
sbetti02: I think there's a typo here that will make it not required (rrequired)

sbetti02 (changes requested):
One small thing I just noticed, also I think we should update the API so it doesn't require the request to send in the user ID, but can figure that out from the request details (logged in user who sent it), if possible

server/routes/events.js
reshadbinharun: @sbetti02 This may cue in a broader backend refactor for how we handle role-based access and authentication. Good eye. We can discuss this during the next standup.

server/models/eventSchema.js
```diff
@@ -4,13 +4,14 @@ const Schema = mongoose.Schema;
...
-        date: {type: Date, rrequired: true},
+        creator: {type: Schema.Types.ObjectId, ref: 'User', required: true},
+        title: {type: String, required: true},
+        date: {type: Date, required: true},
         link: {type: String, required: true},
         description: {type: String, required: true},
         school: {type: Schema.Types.ObjectId, ref: 'School', required: false},
-        years: [{type: Number, required: false}]
+        startYear: {type: Number, required: false},
+        endYear: {type: Number, required: false}
```
sbetti02, line 14: What do we store in the database if we want everyone from the school to be invited?

server/routes/events.js
```diff
@@ -25,74 +25,146 @@ const isValidAccessContext = ({ accessContexts, userSchool, school }) => {
     }
 }
 
+/*
+    Helper function to ensure user has access to a given school
```
sbetti02: Does this comment match what this function does?

server/routes/utilMongoose.js
```diff
@@ -782,6 +827,17 @@ router.get('/data/clear/collegeRep', async (req, res) => {
     }
 })
 
+router.post('/addEvent', async (req, res) => {
+    try {
+        await createEvent(req.body.creator, req.body.title, req.body.date, req.body.link, req.body.description, req.body.school, req.body.startYear, req.body.endYear)
```
sbetti02, line 832: We're still having the creator be included in the API request? I thought we were gonna pull that out from the request info?

server/models/eventSchema.js
focitti: I think the default is use all years unless specifies otherwise because neither `startYear` nor `endYear` are required. This also allows for a bit of flexibility with the year ranges so when creating events, you could also specify you want to include student after x year or student before y year.

server/routes/events.js
focitti: Hmmmm, you have a point sir.

server/routes/utilMongoose.js
focitti: This is for utilMongoose which as far as I know does not include passport auth stuff because it is only really used for seeding/testing. Or maybe I am not fully understanding your question.

server/models/eventSchema.js
sbetti02: Cool, sounds good!

server/routes/utilMongoose.js
sbetti02: Sorry, yeah I might've asked this in the wrong place.. I think what I was referring to was the in the place we submit the request, does the creator (or creator id) need to be specified in the body of the request? Or can that be inferred via the passport auth stuff?

focitti: Oh that makes sense! Yes, the end points were updated so the user information is retrieved from the request without having the FE send it back.

PR #170, opened 2021-03-04
Student/Alumni are awarded footyPoints when users they referred sign up

reshadbinharun:
Alumni and students can now copy a personalized referral link that students and alumni from their own school can use to sign up. The referring alumnus/student is awarded 10 footyPoints on sign up.

Bonus:
- We have React toasts that can be used through the app

Screens:
image
Custom referral links Front and center

Toast to hint that link is copied to clipboard (See in action on test.onefootin.org)

image
Custom link pre-populates school and takes the user to relevant role's sign up page

Manual Testing steps:

* Sign in as a student/alumnus
* Generate a personalized referral link for a student
* Confirm you see the toast and the link is copied to clipboard
* Confirm you are brought to student registration page with the correct school already prepopulated
* Confirm you can sign up
* Repeat for generating link for alumnus

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -25,6 +25,7 @@ import * as actions from './redux/actions'
 import Polls from './components/admin_dashboard/Polls';
 import Footer from './components/Footer'
 import LandingPage from './landingPageContent/LandingPage';
+import ReferralLinkGenerator from './ReferralLinkGenerator';
 
 export const ALUMNI = "ALUMNI"
 export const STUDENT = "STUDENT"
@@ -459,6 +460,11 @@ class App extends Component {
                           timezoneActive={true}
                           navItems={alumniNavBarItems(this.state.approved, this.state.newRequestsCount, this.state.unseenMessagesCount)}
                           activeItem={'home'}
+                      />
+                      <ReferralLinkGenerator
+                        schoolId={this.state.userDetails.school._id}
+                        schoolName={this.state.userDetails.school.name}
+                        userId={this.state.userDetails.user}
                       />
                       <NewsFeed
                         userDetails={this.state.userDetails}
@@ -969,7 +975,18 @@ class App extends Component {
                   this.state.loggedIn ? <Redirect to={"/"}/> : 
                   <Signup
                       isAlumni={true}
-                      match={props}
+                      match={props.match}
+                      history={props.history}
+                  />
+              }
+          />
+          <Route exact path={`/register/alumni/:referrerId/:schoolId`} render={
+              (props) =>
+                  this.state.loggedIn ? <Redirect to={"/"}/> : 
+                  <Signup
+                      isAlumni={true}
+                      match={props.match}
+                      history={props.history}
                   />
               }
           />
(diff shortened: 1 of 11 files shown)


server/routes/alumni.js
```diff
@@ -20,9 +20,26 @@ const opportunitySchema = require('../models/opportunitySchema');
...
 require('mongoose').Promise = global.Promise
+var FOOTY_POINTS_CHART = require('../footyPointsChart').FOOTY_POINTS_CHART
 
 const HASH_COST = 10;
 
+/**
+ * Finds the alumnus or student account by the id of the referrer provided and awards footyPoints
+ * @param {*} referrerId 
+ */
+const awardReferralPoints = async (referrerId) => {
```
focitti, line 31: nit: it might be worth having a footyPoint helper file or something along those lines. Because as the concept grows, figuring out where methods like this belong may be a little harder.

client/src/App.js
```diff
@@ -969,17 +976,39 @@ class App extends Component {
                   this.state.loggedIn ? <Redirect to={"/"}/> : 
                   <Signup
                       isAlumni={true}
-                      match={props}
+                      match={props.match}
+                      history={props.history}
+                  />
+              }
+          />
+          <Route exact path={`/register/alumni/:referrerId/:schoolId`} render={
```
focitti, line 984: nit beyond scope of this PR: it would be nice to consolidate the logged in logic check
question within the scope of this PR: why does the register/alumni path have to have two block cases here (i.e. why use and exact match on the route in this case)? It looks like they are both taking in same props.

client/src/ReferralLinkGenerator.js
```diff
@@ -0,0 +1,59 @@
...
+import { toast } from 'react-semantic-toasts'
+
+/**
+ * Button group that allows a user to copy a referral link to share
+ * @param {*} props
+ * schoolId - id for the user current school
+ * schoolName - name of the user school
+ * userId - userId for the student / alumnus 
+ */
+export default function ReferralLinkGenerator (props) {
```
focitti, line 12: perhaps a future suggestion, but it would be nice to maybe have a modal where the user can enter the email of the person they are referring. That way we can keep activation energy lower by keeping it in app.

client/src/ReferralLinkGenerator.js
```diff
@@ -0,0 +1,59 @@
...
+                    description: <p>Your {role.toUpperCase()} referral link has been copied to clipboard!</p>,
+                    color: role === 'alumni' ? 'orange' : 'yellow',
+                    time: 2000,
+                    icon: 'paper plane',
+                }
+            );
+        })
+    }
+
+    return (
```
focitti, line 29: also do not quite have developed thoughts on this yet, but the card is not quite distinctive. When reviewing the change, I did not immediately recognise where the refer button was. Things that may help:
* Maybe inverting the colours? And making the buttons solid coloured. That way the they are more apparent and do not blend it with the app background.
* Inversing the order s.t. it is button above, descriptive text below. Because there is no title in the card, I think it would help with app flow if the first thing the user saw was the action instead of the description of the action. If more context is needed then the user will go on to read the description.

client/src/components/Signup.js
```diff
@@ -438,6 +439,15 @@ export default class Signup extends React.Component {
     }
 
     async UNSAFE_componentWillMount() {
```
focitti, line 441: why the use of `UNSAFE_componentWillMount`? I looked it up to see the diff between the UNSAFE version and the other version and it appears this one is legacy code.

focitti (changes requested):
This looks great! Main comment just about the UI of the link generator card.

